PR #266, opened 2022-05-30
[Fix] Fix ema state dict swapping in EMAHook and torch1.5 ut.

RangiLyu:
# 1. Fix state dict swapping
## Motivation

The original implementation of `_swap_ema_parameters` can not swap the weight in the checkpoint, because the state_dict is a ***shallow copy*** of model weights.

## Modification

Add `_swap_ema_state_dict` to swap the weights in the checkpoint.

# 2. Fix torch1.5 unit tests
## Motivation
the return tensor' data of the `detach` operation in 1.5 does not point to the same memory address so the inplace-operation can not modify the origin model's parameters in the unit tests.

## Modification

Replace `detach().add_()` with `add()`

### DIFF
--- a/mmengine/hooks/ema_hook.py
+++ b/mmengine/hooks/ema_hook.py
@@ -71,8 +71,11 @@ class EMAHook(Hook):
     def before_save_checkpoint(self, runner, checkpoint: dict) -> None:
         """Save ema parameters to checkpoint."""
         checkpoint['ema_state_dict'] = self.ema_model.state_dict()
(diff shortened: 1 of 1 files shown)


ykhf:
你好，想问下，‘The original implementation of _swap_ema_parameters can not swap the weight in the checkpoint, because the state_dict is a shallow copy of model weights.'  这个会导致训练模型时一直没有更新到权重吗？ 训练完成的save_checkpoint 保存的模型也是错误的模型吗？

PR #364, opened 2022-07-14
[Fix] Fix type assert in build_model of runner

Yshuo-Li:
## Motivation

Fix type assert in build_model of runner

## Modification

``mmengine/runner/runner.py::Runner.build_model``

from ``BaseModel`` to ``nn.Module``


HAOCHENYE:
Fixed in https://github.com/open-mmlab/mmengine/commit/2df5bc137ba82c6af31d6f25b07297b7db695ac1

PR #533, opened 2022-09-14
[Refactor] Refactor fileio but without breaking bc

zhouzaida:

## Motivation

In this PR, we refactor the **fileio** module to make it more user-friendly and provide more frequently used functions. Note that this PR will not cause backward compatibility breaking which means the **previous case** works fine until the release of **MMEngine v1.0**.

- Previous Case

```python
>>> from mmengine.fileio import FileClient
>>> client = FileClient()
>>> client.get_text('s3://path/of/your/file')
'hello world'
```

- New Case

```python
>>> import mmengine.fileio as fileio
>>> fileio.get_text('s3://path/of/your/file')
'hello world'
```

## Modification

1. Move the file backends in `mmengine/fileio/file_client.py` to `mmengine/fileio/backends` but import them in former to avoide BC-breaking. Besides, rename `HardDiskBackend` to `LocalBackend` because the former class name is not enough accuracy. To avoid bc-breaking, still define `HardDiskBackend` in `mmengine/fileio/file_client.py` which inherits from `LocalBackend` and does not have any modifications. Each file contains a `Backend`.

    - mmengine/fileio/backends/base.py
    - mmengine/fileio/backends/local_backend.py (LocalBackend)
    - mmengine/fileio/backends/petrel_backend.py (PetrelBackend)
    - mmengine/fileio/backends/memcached_backend.py (MemcachedBackend)
    - mmengine/fileio/backends/lmdb_backend.py (LmdbBackend)
    - mmengine/fileio/backends/http_backend.py (HTTPBackend)

2. Provide more interfaces for File Backends
    
    - get
    - get_text
    - put
    - put_text
    - exists
    - isdir
    - isfile
    - join_path
    - get_local_path
    - copyfile (new method)
    - copytree( new method)
    - copyfile_from_local (new method)
    - copytree_from_local (new method)
    - copyfile_to_local (new method)
    - copytree_to_local (new method)
    - remove
    - rmtree (new method)
    - copy_if_symlink_fails (new method)
    - list_dir_of_file

   **Note: Only LocalBackend and PetrelBackend implement all of the above interfaces**.

3. Functions or methods involving FileClient have also been modified

    - mmengine/hooks/checkpoint_hook.py
   
      Add a new argument `backend_args` to the construct method of `CheckpointHook`. The `backend_args` is a dictionary which probably contains a `backend` key-value determining the backend that will be used , and the rest of it will be passed to the `__init__` method of backends.
      The difference between the filelds of `file_client_args` and `backend_args` is that `file_client_args` accepts the prefix field, while `backend_args` does not.

      ```python
      # passing `file_client_args` will print a deprecated warning but still works
      hook = CheckpointHook(xxx, file_client_args={'backend': 'disk'})
      # passing `backend_args` will work fine
      # the value of `backend` should be `local` mapping to 'LocalBackend' rather than `disk`
      hook = CheckpointHook(xxx, backend_args={'backend': 'local'})
      ```

    - mmengine/hooks/logger_hook.py

      same as above

    - mmengine/runner/checkpoint.py
    - mmengine/runner/runner.py
      Add a new argument `backend_args` to the `save_checkpoint` method of `Runner`.

## BC-breaking

There is no bc-breaking util release of **MMEngine v1.0.0** (current version is v0.1.0)

## Use cases

There are two ways to call a method of a file backend:

- Initialize a file backend with ``get_file_backend`` and call its methods.
- Directly call unified I/O functions, which will call ``get_file_backend``
  first and then call the corresponding backend method.

```python
>>> # Initialize a file backend and call its methods
>>> import mmengine.fileio as fileio
>>> backend = fileio.get_file_backend(backend_args={'backend': 'petrel'})
>>> backend.get('s3://path/of/your/file')
b'hello world'
>>> # use the global backend instance
>>> backend1 = fileio.get_file_backend(backend_args={'backend': 'petrel'}, enable_singleton=True)
>>> backend is backend1
True

>>> # Directly call unified I/O functions and use the global backend instance by default
>>> fileio.get('s3://path/of/your/file')
b'hello world'
```

## Validate the modification in downstream repos

Validate whether the old version and current version works fine as expected in [MMDetection](https://github.com/open-mmlab/mmdetection/tree/dev-3.x).

**Read training data from Petrel**

- Old version (Tested)
 
```python
file_client_args = dict(
    backend='petrel',
    path_mapping=dict({
        './data/': 's3://openmmlab/datasets/detection/',
        'data/': 's3://openmmlab/datasets/detection/'})
)
```
 - New Version

This new version requires [mmcv](https://github.com/open-mmlab/mmcv/blob/2a82b52b8d4a0d75d133ffdabf894ef4af39fb3d/mmcv/transforms/loading.py#L36) and [mmdet](https://github.com/open-mmlab/mmdetection/blob/af4604f9dc6f45b35305160fad91fa5a6e8ae382/mmdet/datasets/base_det_dataset.py#L26) support accepting `backend_args` argument so this new version will not be tested.

```python
backend_args = dict(
    backend='petrel',
    path_mapping=dict({
        './data/': 's3://openmmlab/datasets/detection/',
        'data/': 's3://openmmlab/datasets/detection/'
  }))
```

**Load or resume checkpoints from Petrel** (Tested)

```python
load_from = 'myself:s3://zhouzaida/mmdet/xxxx.ckpt'
```

**Save checkpoints to Petrel**

- Old version (Tested)

```python
default_hooks = dict(
    checkpoint=dict(type='CheckpointHook', interval=1, max_keep_ckpts=2, out_dir='myself:s3://zhouzaida/mmdet/', file_client_args={'backend': 'petrel'}),
)
```

- New version (Tested)

```python
default_hooks = dict(
    checkpoint=dict(type='CheckpointHook', interval=1, max_keep_ckpts=2, out_dir='myself:s3://zhouzaida/mmdet/', backend_args={'backend': 'petrel'}),
)
```

**Save logs to Petrel**

- Old version (Tested)

default_hooks = dict(
    logger=dict(type='LoggerHook', interval=50, out_dir='myself:s3://zhouzaida/mmdet/', file_client_args={'backend': 'petrel'}),
)

- New version (Tested)

```python
default_hooks = dict(
    logger=dict(type='LoggerHook', interval=50, out_dir='myself:s3://zhouzaida/mmdet/', backend_args={'backend': 'petrel'}),
)
```

## Checklist

4. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
7. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
8. The documentation has been modified accordingly, like docstring or example tutorials.


mmengine/fileio/backends/http_backend.py
```diff
@@ -0,0 +1,74 @@
...
+from urllib.request import urlopen
+
+from .base import BaseStorageBackend
+
+
+class HTTPBackend(BaseStorageBackend):
+    """HTTP and HTTPS storage bachend."""
+
+    def get_bytes(self, filepath: str) -> bytes:
+        """ead bytes from a given ``filepath``.
```
C1rN09, line 16:
```suggestion
        """read bytes from a given ``filepath``.
```

mmengine/fileio/backends/http_backend.py
```diff
@@ -0,0 +1,74 @@
...
+        Args:
+            filepath (str): Download a file from ``filepath``.
+
+        Examples:
+            >>> backend = HTTPBackend()
+            >>> # After existing from the ``with`` clause,
+            >>> # the path will be removed
+            >>> with backend.get_local_path('http://path/of/file') as path:
+            ...     # do something here
+        """
```
C1rN09, line 67: The function name and description is kind of misleading to me, probably because it's doing multiple works.
Might be better if description modified as:
```
Download a file from ``filepath`` to a local temporary directory, and return the local path.
```
and add the return value description.

mmengine/fileio/io.py
```diff
@@ -15,7 +52,735 @@
...
+    if uri is not None:
+        prefix = _parse_uri_prefix(uri)
+    else:
+        prefix = ''
+
+    if enable_singleton:
+        # TODO: whether to pass sort_key to json.dumps
+        unique_key = f'{prefix}:{json.dumps(backend_args)}'
+        if unique_key in backend_instances:
+            return backend_instances[unique_key]
```
C1rN09, line 152: I think it's good to pass `sort_key` to `json.dumps`

mmengine/fileio/backends/lmdb_backend.py
```diff
@@ -0,0 +1,80 @@
...
+
+        return lmdb.open(
+            self.db_path,
+            readonly=self.readonly,
+            lock=self.lock,
+            readahead=self.readahead,
+            **self.kwargs)
+
+    def __del__(self):
+        self._client.close()
```
C1rN09, line 80: current use cases are fine, but is there any possibility that `LmdbBackend` is initialized without calling `get_xxx` methods?
I think it's better to:
1. initialize `self._client` at `__init__` rather than `get_xxx`
2. or, check whether `self._client` is `None` here

mmengine/hooks/logger_hook.py
```diff
@@ -91,9 +106,16 @@ def __init__(self,
...
+
         if self.out_dir is not None:
             self.file_client = FileClient.infer_client(file_client_args,
                                                        self.out_dir)
+            if file_client_args is None:
+                self.file_backend = get_file_backend(
+                    self.out_dir, backend_args=backend_args)
+            else:
+                self.file_backend = self.file_client
+
```
C1rN09, line 118:
```suggestion
        if self.out_dir is not None:
            if file_client_args is None:
                self.file_backend = get_file_backend(
                    self.out_dir, backend_args=backend_args)
            else:
                self.file_backend = FileClient.infer_client(file_client_args,
                                                                                      self.out_dir)

```

mmengine/fileio/backends/local_backend.py
```diff
@@ -0,0 +1,522 @@
...
+
+        Yields:
+            Iterable[str]: A relative path to ``dir_path``.
+
+        Examples:
+            >>> backend = LocalBackend()
+            >>> dir_path = '/path/of/dir'
+            >>> for file_path in backend.list_dir_or_file(dir_path):
+            ...     print(file_path)
+        """
```
C1rN09, line 495: This method is complicated. I think more examples are needed.

C1rN09 (commented):
Basically good.
Will `xxxBackend` codes in `file_client.py` be removed in the future, since they are moved to `backends` directory?

zhouzaida:
> Basically good. Will `xxxBackend` codes in `file_client.py` be removed in the future, since they are moved to `backends` directory?

They had been moved in the newest commit.

mmengine/fileio/backends/lmdb_backend.py
zhouzaida: More details at https://github.com/open-mmlab/mmcv/pull/1764.

mmengine/hooks/logger_hook.py
zhouzaida: Keep `self.file_client_args` to avoid bc-breaking. This property is basically not accessed by the caller, so consider removing it to simplify the code.

mmengine/fileio/io.py
zhouzaida: If `backend_args` contains string key and integer key, passing `sort_keys` to `json.dumps` will throw a `TypeError`.

```python
>>> import json
>>> json.dumps(a, sort_keys=True)
'[11, 222]'
>>> a = {'a1': [11], 1: [13]}
>>> json.dumps(a, sort_keys=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/developer/miniconda3/envs/py37-pt120-cu116/lib/python3.7/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/home/developer/miniconda3/envs/py37-pt120-cu116/lib/python3.7/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/home/developer/miniconda3/envs/py37-pt120-cu116/lib/python3.7/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
TypeError: '<' not supported between instances of 'int' and 'str'
>>> json.dumps(a)
'{"a1": [11], "1": [13]}'
```

C1rN09: Current implementation will create multiple instances if arguments are in different orders.
e.g. `args1 = {'k1': 1, 'k2': 2}` and `args2 = {'k2': 2, 'k1': 1}`
Maybe we should work out a method to sort `type` first and then `elements` with the same type. Is builtin function `id` feasible?

mmengine/fileio/backends/lmdb_backend.py
C1rN09: I see, so it's necessary to initialize it in `getxxx` method. Then I think it's better to check for `None` in `__del__`
```python
def __del__(self):
    if self._client: self._client.close()
```

zhouzaida: Great suggestion

mmengine/fileio/backends/local_backend.py
zhouzaida: updated

mmengine/fileio/io.py
zhouzaida: The reason to use singleton is to avoid creating a lot of instances repeatedly but it is acceptable to create multiple instances. Therefore, it may not be necessary to handle it this way.

mmengine/fileio/backends/memcached_backend.py
```diff
@@ -0,0 +1,43 @@
...
+
+        self.server_list_cfg = server_list_cfg
+        self.client_cfg = client_cfg
+        self._client = mc.MemcachedClient.GetInstance(self.server_list_cfg,
+                                                      self.client_cfg)
+        # mc.pyvector servers as a point which points to a memory cache
+        self._mc_buffer = mc.pyvector()
+
+    def get(self, filepath: Union[str, Path]):
+        filepath = str(filepath)
```
C1rN09, line 36: docstring missing

zhouzaida: done

C1rN09 (dismissed):

mmengine/fileio/backends/lmdb_backend.py
```diff
@@ -0,0 +1,81 @@
...
+    Args:
+        db_path (str): Lmdb database path.
+        readonly (bool): Lmdb environment parameter. If True, disallow any
+            write operations. Defaults to True.
+        lock (bool): Lmdb environment parameter. If False, when concurrent
+            access occurs, do not lock the database. Defaults to False.
+        readahead (bool): Lmdb environment parameter. If False, disable the OS
+            filesystem readahead mechanism, which may improve random read
+            performance when a database is larger than RAM. Defaults to False.
+
```
ZwwWayne, line 20: We may also indicate the kwargs and tell users that they will be sent to build the lmdb.open

mmengine/fileio/io.py
```diff
@@ -102,48 +904,33 @@ def dump(obj, file=None, file_format=None, file_client_args=None, **kwargs):
...
+            raise ValueError(
+                '"file_client_args and "backend_args" cannot be both set.')
+
     handler = file_handlers[file_format]
     if file is None:
         return handler.dump_to_str(obj, **kwargs)
     elif is_str(file):
-        file_client = FileClient.infer_client(file_client_args, file)
+        if file_client_args is not None:
+            file_client = FileClient.infer_client(file_client_args, file)
```
ZwwWayne, line 920: does file_client_args have the same key values as those in backend_args?
If they have, we may consider using an existing function named deprecation_warning to directly map the file_client_args to backend_args.

ZwwWayne: If not, we should put backend_args before file_client_args since it is an argument to be deprecated.

mmengine/runner/runner.py
```diff
@@ -2047,8 +2055,19 @@ def save_checkpoint(self,
...
 
-        file_client = FileClient.infer_client(file_client_args, out_dir)
-        filepath = file_client.join_path(out_dir, filename)
+        if file_client_args is not None:
+            warnings.warn(
+                '"file_client_args" will be deprecated in future. '
+                'Please use "backend_args" instead', DeprecationWarning)
+            if backend_args is not None:
+                raise ValueError(
+                    '"file_client_args and "backend_args" cannot be both set.')
```
ZwwWayne, line 2064: cannot be set at the same time.

mmengine/fileio/io.py
zhouzaida: > does file_client_args have the same key values as those in backend_args? If they have, we may consider using an existing function named deprecation_warning to directly map the file_client_args to backend_args.

file_client_args has an extra key 'prefix'

zhouzaida: > If not, we should put backend_args before file_client_args since it is an argument to be deprecated.

Changing the order of parameters is to cause bc-breaking.

ZwwWayne (approved):

PR #549, opened 2022-09-26
[Fix] Use `get` instead of `pop` to dump `runner_type`

nijkah:
## Motivation
When using a custom runner, `runner_type` is not dumped appropriately, since `build_runner_from_cfg` function **pops** the `runner_type`.
To reproduce the experiment, it seems better to dump `runner_type` too.
https://github.com/open-mmlab/mmengine/blob/8d14eb65465785f8ec45bd8125a39c918f601b1f/mmengine/registry/build_functions.py#L179


HAOCHENYE:
LGTM.

HAOCHENYE (approved):

PR #551, opened 2022-09-27
[Feature] Support torch ZeroRedundancyOptimizer

nijkah:
- Device: V100 32GB (32510MiB) x4
- Input Size: 512
- Total Batch Size: 16

### upernet_r50_4xb4-80k_ade20k-512x512

|  Method  |      Time       |  Memory  |
| :------: | :-------------: | :------: |
| w/o zero | 0.611 | 7053|
| w/  zero | 0.552 | 6868 |

Time may be dependent to the environment.

Co-authored-by: Junhwa Song <ethan9867@gmail.com> @KKIEEK
Signed-off-by: Junhwa Song <ethan9867@gmail.com>
Signed-off-by: Hakjin Lee <nijkah@gmail.com>

- [x] Check reduced memory
- [x] Reproduce Performance
- [x] Check saving optimizer's state
- [x] Check loading optimizer's state

### DIFF
--- a/mmengine/optim/optimizer/zero_optimizer.py
+++ b/mmengine/optim/optimizer/zero_optimizer.py
@@ -57,8 +57,7 @@ class ZeroRedundancyOptimizer(_ZeroRedundancyOptimizer):
         # TODO: Register a DDP communication hook for `overlap_with_ddp=True`.
         # Currently only `overlap_with_ddp=False` is supported. For more
         # details, please refer to the pytorch's official documentation.
-        super().__init__(
-            params, optimizer_class, overlap_with_ddp=False, **kwargs)
+        super().__init__(params, optimizer_class, **kwargs)
 
     def state_dict(self):
         """Consolidate `state_dict`s from ranks to save the `state_dict`."""
--- a/mmengine/optim/optimizer/zero_optimizer.py
+++ b/mmengine/optim/optimizer/zero_optimizer.py
@@ -24,12 +24,12 @@ class ZeroRedundancyOptimizer(_ZeroRedundancyOptimizer):
     This class wraps an arbitrary :class:`torch.optim.Optimizer` and shards its
     states across ranks in the group as described by ZeRO_. The local optimizer
     instance in each rank is only responsible for updating approximately
-    ``1 / world_size`` parameters and hence only needs to keep ``1 / world_size``
-    optimizer states. After parameters are updated locally, each rank will
-    broadcast its parameters to all other peers to keep all model replicas
-    in the same state. ``ZeroRedundancyOptimizer`` can be used in conjunction
-    with :class:`torch.nn.parallel.DistributedDataParallel` to reduce per-rank peak
-    memory consumption.
+    ``1 / world_size`` parameters and hence only needs to keep
+    ``1 / world_size`` optimizer states. After parameters are updated locally,
+    each rank will broadcast its parameters to all other peers to keep all
+    model replicas in the same state. ``ZeroRedundancyOptimizer`` can be used
+    in conjunction with :class:`torch.nn.parallel.DistributedDataParallel` to
+    reduce per-rank peak memory consumption.
 
     ``ZeroRedundancyOptimizer`` uses a sorted-greedy algorithm to pack a number
     of parameters at each rank. Each parameter belongs to a single rank and is
--- a/mmengine/optim/optimizer/zero_optimizer.py
+++ b/mmengine/optim/optimizer/zero_optimizer.py
@@ -19,29 +19,33 @@ from .builder import OPTIMIZERS
 @OPTIMIZERS.register_module()
 class ZeroRedundancyOptimizer(_ZeroRedundancyOptimizer):
     """A wrapper class of :class:`ZeroRedundancyOptimizer` that gets a
-    optimizer type as string. This class wraps an arbitrary
-    :class:`optim.Optimizer.
-
-    <torch.optim.Optimizer>` and shards its states across ranks in the group as
-    described by ZeRO_. The local optimizer instance in each rank is only
-    responsible for updating approximately ``1 / world_size`` parameters and
-    hence only needs to keep ``1 / world_size`` optimizer states. After
-    parameters are updated locally, each rank will broadcast its parameters to
-    all other peers to keep all model replicas in the same state.
-    ``ZeroRedundancyOptimizer`` can be used in conjunction with
-    :class:`torch.nn.parallel.DistributedDataParallel` to reduce per-rank peak
+    optimizer type as string. 
+    
+    This class wraps an arbitrary :class:`torch.optim.Optimizer` and shards its
+    states across ranks in the group as described by ZeRO_. The local optimizer
+    instance in each rank is only responsible for updating approximately
+    ``1 / world_size`` parameters and hence only needs to keep ``1 / world_size``
+    optimizer states. After parameters are updated locally, each rank will
+    broadcast its parameters to all other peers to keep all model replicas
+    in the same state. ``ZeroRedundancyOptimizer`` can be used in conjunction
+    with :class:`torch.nn.parallel.DistributedDataParallel` to reduce per-rank peak
     memory consumption.
+
     ``ZeroRedundancyOptimizer`` uses a sorted-greedy algorithm to pack a number
     of parameters at each rank. Each parameter belongs to a single rank and is
     not divided among ranks. The partition is arbitrary and might not match the
     the parameter registration or usage order.
+
     Warnings:
         ``ZeroRedundancyOptimizer`` requires PyTorch >= 1.8.
+
     Args:
         params (``Iterable``): an ``Iterable`` of :class:`torch.Tensor` s
             or :class:`dict` s giving all parameters, which will be sharded
             across ranks.
         optimizer_type (str): the string of the local optimizer class.
+        
+    .. _ZeRO: https://arxiv.org/abs/1910.02054
     """
 
     def __init__(self, params, optimizer_type: str, **kwargs):
--- a/mmengine/optim/optimizer/zero_optimizer.py
+++ b/mmengine/optim/optimizer/zero_optimizer.py
@@ -59,5 +59,5 @@ class ZeroRedundancyOptimizer(_ZeroRedundancyOptimizer):
     def state_dict(self):
         """Consolidate `state_dict`s from ranks to save the `state_dict`."""
         self.consolidate_state_dict()
-        if is_main_process():
-            return super().state_dict()
+        state_dict = super().state_dict() if is_main_process() else dict()
+        return state_dict
--- a/mmengine/optim/optimizer/zero_optimizer.py
+++ b/mmengine/optim/optimizer/zero_optimizer.py
@@ -11,7 +11,7 @@ try:
     from torch.distributed.optim import \
         ZeroRedundancyOptimizer as _ZeroRedundancyOptimizer
 except ImportError:
-    _ZeroReundancyOptimizer = object
+    _ZeroRedundancyOptimizer = object
 
 from .builder import OPTIMIZERS
 
--- a/mmengine/optim/optimizer/zero_optimizer.py
+++ b/mmengine/optim/optimizer/zero_optimizer.py
@@ -9,7 +9,7 @@ from mmengine.utils.dl_utils import TORCH_VERSION
 
 try:
     from torch.distributed.optim import \
-        ZeroRedundancyOptimizer as _ZeroReundancyOptimizer
+        ZeroRedundancyOptimizer as _ZeroRedundancyOptimizer
 except ImportError:
     _ZeroReundancyOptimizer = object
 
@@ -17,7 +17,7 @@ from .builder import OPTIMIZERS
 
 
 @OPTIMIZERS.register_module()
-class ZeroRedundancyOptimizer(_ZeroReundancyOptimizer):
+class ZeroRedundancyOptimizer(_ZeroRedundancyOptimizer):
     """A wrapper class of :class:`ZeroRedundancyOptimizer` that gets a
     optimizer type as string. This class wraps an arbitrary
     :class:`optim.Optimizer.
--- a/mmengine/optim/optimizer/zero_optimizer.py
+++ b/mmengine/optim/optimizer/zero_optimizer.py
@@ -5,12 +5,13 @@ import torch
 from mmengine.dist import is_main_process
 from mmengine.utils import digit_version
 from mmengine.utils.dl_utils import TORCH_VERSION
+import warnings
 
 try:
     from torch.distributed.optim import \
         ZeroRedundancyOptimizer as _ZeroReundancyOptimizer
 except ImportError as e:
-    print(e)
+    warnings.warn(e)
     _ZeroReundancyOptimizer = object
 
 from .builder import OPTIMIZERS
--- a/mmengine/optim/optimizer/zero_optimizer.py
+++ b/mmengine/optim/optimizer/zero_optimizer.py
@@ -2,13 +2,15 @@
 
 import torch
 
+from mmengine.dist import is_main_process
 from mmengine.utils import digit_version
 from mmengine.utils.dl_utils import TORCH_VERSION
 
 try:
     from torch.distributed.optim import \
         ZeroRedundancyOptimizer as _ZeroReundancyOptimizer
-except ImportError:
+except ImportError as e:
+    print(e)
     _ZeroReundancyOptimizer = object
 
 from .builder import OPTIMIZERS
@@ -52,4 +54,5 @@ class ZeroRedundancyOptimizer(_ZeroReundancyOptimizer):
     def state_dict(self):
         """Consolidate `state_dict`s from ranks to save the `state_dict`"""
         self.consolidate_state_dict()
-        return super().state_dict()
+        if is_main_process():
+            return super().state_dict()
--- a/mmengine/optim/optimizer/zero_optimizer.py
+++ b/mmengine/optim/optimizer/zero_optimizer.py
@@ -9,7 +9,7 @@ try:
     from torch.distributed.optim import \
         ZeroRedundancyOptimizer as _ZeroReundancyOptimizer
 except ImportError:
-    _ZeroReundancyOptimizer = None
+    _ZeroReundancyOptimizer = object
 
 from .builder import OPTIMIZERS
 
--- a/mmengine/optim/optimizer/zero_optimizer.py
+++ b/mmengine/optim/optimizer/zero_optimizer.py
@@ -1,0 +1,48 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+
+import torch
+from mmengine.utils import digit_version
+from mmengine.utils.dl_utils import TORCH_VERSION
+
+try:
+    from torch.distributed.optim import \
+        ZeroRedundancyOptimizer as _ZeroReundancyOptimizer
+except ImportError:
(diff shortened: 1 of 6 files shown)


nijkah:
```
  File "/workspace/R-BIG22A/mmengine/mmengine/runner/runner.py", line 1718, in call_hook                                                                                                                                          
    getattr(hook, fn_name)(self, **kwargs)                                                                                                                                                                                        
  File "/workspace/R-BIG22A/mmengine/mmengine/hooks/checkpoint_hook.py", line 488, in after_train_iter
    self._save_checkpoint(runner)                                                                                                                                                                                                 
  File "/workspace/R-BIG22A/mmengine/mmengine/dist/utils.py", line 346, in wrapper                                                                                                                                                
    return func(*args, **kwargs)                                                                                                                                                                                                  
  File "/workspace/R-BIG22A/mmengine/mmengine/hooks/checkpoint_hook.py", line 294, in _save_checkpoint                                                                                                                            
    **self.args)                                                                                                                                                                                                                  
  File "/workspace/R-BIG22A/mmengine/mmengine/runner/runner.py", line 2109, in save_checkpoint                                                                                                                                    
    ) else self.optim_wrapper.state_dict()                                                                                                                                                                                        
  File "/workspace/R-BIG22A/mmengine/mmengine/optim/optimizer/optimizer_wrapper.py", line 199, in state_dict                                                                                                                      
    return self.optimizer.state_dict()                                                                                                                                                                                            
  File "/opt/conda/lib/python3.7/site-packages/torch/distributed/optim/zero_redundancy_optimizer.py", line 1152, in state_dict                                                                                                    
    "Optimizer state has not been consolidated on this rank. "                                                                                                                                                                    
RuntimeError: Optimizer state has not been consolidated on this rank. Please call `consolidate_state_dict(to=0)` on all ranks beforehand if you meant to save the global state. 
```

I found that it has a problem when saving the optimizer's state_dict.

nijkah:
```
  File "/workspace/mmengine/mmengine/runner/runner.py", line 1718, in call_hook                                                                                                                                          
    getattr(hook, fn_name)(self, **kwargs)                                                                                                                                                                                        
  File "/workspace/mmengine/mmengine/hooks/checkpoint_hook.py", line 488, in after_train_iter
    self._save_checkpoint(runner)                                                                                                                                                                                                 
  File "/workspace/mmengine/mmengine/dist/utils.py", line 346, in wrapper                                                                                                                                                
    return func(*args, **kwargs)                                                                                                                                                                                                  
  File "/workspace/mmengine/mmengine/hooks/checkpoint_hook.py", line 294, in _save_checkpoint                                                                                                                            
    **self.args)                                                                                                                                                                                                                  
  File "/workspace/mmengine/mmengine/runner/runner.py", line 2109, in save_checkpoint                                                                                                                                    
    ) else self.optim_wrapper.state_dict()                                                                                                                                                                                        
  File "/workspace/mmengine/mmengine/optim/optimizer/optimizer_wrapper.py", line 199, in state_dict                                                                                                                      
    return self.optimizer.state_dict()                                                                                                                                                                                            
  File "/opt/conda/lib/python3.7/site-packages/torch/distributed/optim/zero_redundancy_optimizer.py", line 1152, in state_dict                                                                                                    
    "Optimizer state has not been consolidated on this rank. "                                                                                                                                                                    
RuntimeError: Optimizer state has not been consolidated on this rank. Please call `consolidate_state_dict(to=0)` on all ranks beforehand if you meant to save the global state. 
```

I found that it has a problem when saving the optimizer's state_dict.

nijkah:
Solved the problem to call `self.state_dict`.

HAOCHENYE:
Thank you for your contributions! It could be better to add a unit test to use the `ZeroRedundancyOptimizer` :blush:.

tests/test_optim/test_optimizer/test_optimizer.py
```diff
@@ -713,3 +722,78 @@ def test_default_optimizer_constructor_custom_key(self):
...
+
+
+@unittest.skipIf(
+    digit_version(TORCH_VERSION) < digit_version('1.8.0'),
+    reason='ZeRO needs Pytorch 1.8 or higher')
+class TestZeroOptimizer(MultiProcessTestCase):
+
+    def setUp(self) -> None:
+        if ZeroRedundancyOptimizer is None:
+            self.skipTest('ZeroRedundancyOptimizer is not available.')
```
HAOCHENYE, line 734:
```suggestion
        if ZeroRedundancyOptimizer is None:
            self.skipTest('ZeroRedundancyOptimizer is not available.')
```

Is this line duplicated with 
```python
@unittest.skipIf(
    digit_version(TORCH_VERSION) < digit_version('1.8.0'),
    reason='ZeRO needs Pytorch 1.8 or higher')
```

nijkah: https://github.com/open-mmlab/mmengine/actions/runs/3134972777/jobs/5090129146#step:8:132
I found that importing `ZeroRedundancyOptimizer` failed in the Windows CPU CI with & `torch1.8.1`.
(Importing failure made `_ZeroRedundancyOptimizer` as object.)
So I added duplicated skip code.

I'll check it again.

tests/test_optim/test_optimizer/test_optimizer.py
```diff
@@ -713,3 +722,78 @@ def test_default_optimizer_constructor_custom_key(self):
...
+
+        super().setUp()
+        self._spawn_processes()
+
+    def _check_default_optimizer(self, optimizer, model, prefix=''):
+        assert isinstance(optimizer.optim, torch.optim.SGD)
+        assert optimizer.defaults['lr'] == self.base_lr
+        assert optimizer.defaults['momentum'] == self.momentum
+        assert optimizer.defaults['weight_decay'] == self.base_wd
```
HAOCHENYE, line 743: Since we inherit `TestCase` defined in unittest, it could be better to use `self.assertIsinstance` and `self.assertEqual`

tests/test_optim/test_optimizer/test_optimizer.py
```diff
@@ -713,3 +722,78 @@ def test_default_optimizer_constructor_custom_key(self):
...
+class TestZeroOptimizer(MultiProcessTestCase):
+
+    def setUp(self) -> None:
+        if ZeroRedundancyOptimizer is None:
+            self.skipTest('ZeroRedundancyOptimizer is not available.')
+
+        super().setUp()
+        self._spawn_processes()
+
+    def _check_default_optimizer(self, optimizer, model, prefix=''):
```
HAOCHENYE, line 739: Is the `prefix` argument necessary?

nijkah:
I found another bug.

Currently, saving the optimizer's `state_dict` gives the error.
Each optimizer should call `consolidate_state_dict` for all processes, and only the main process's optimizer should return the `state_dict`.

This constraint makes refactor the 'Checkpoint Hook.

nijkah:
@HAOCHENYE @C1rN09 

I changed the `CheckpointHook`s logic by removing `master_only` decorators when saving the checkpoint.
I think it doesn't affect the logic since `save_checkpoint` function in `mmengine/runner/checkpoint.py` is wrapped by `master_only` already. 

Please notify me if the potential problem is expected.

tests/test_optim/test_optimizer/test_optimizer.py
nijkah: Deleted.

tests/test_optim/test_optimizer/test_optimizer.py
nijkah: Fixed.

tests/test_optim/test_optimizer/test_optimizer.py
nijkah: I found that it has another condition.
`torch.distributed.rpc` should be available. I removed duplicated lines, and clarified this condition.

HAOCHENYE:
#

> Hi, @HAOCHENYE @C1rN09
> 
> I changed the `CheckpointHook`'s logic to remove `master_only` decorators when saving the checkpoint. I think it doesn't affect the logic since `save_checkpoint` function in `mmengine/runner/checkpoint.py` is wrapped by `master_only` already.
> 
> Please notify me if the potential problem is expected.

#553 also remove the `master_only` decorator to collect all weights from different rank. Does @C1rN09 have any idea about this?

mmengine/optim/optimizer/zero_optimizer.py
```diff
@@ -0,0 +1,59 @@
...
+        params (``Iterable``): an ``Iterable`` of :class:`torch.Tensor` s
+            or :class:`dict` s giving all parameters, which will be sharded
+            across ranks.
+        optimizer_type (str): the string of the local optimizer class.
+    """
+
+    def __init__(self, params, optimizer_type: str, **kwargs):
+        assert digit_version(TORCH_VERSION) >= digit_version('1.8.0'), (
+            '`torch.distributed.optim.ZeroReundancyOptimizer` is only '
+            'available when pytorch version >= 1.8')
```
nijkah, line 50:
```suggestion
            'available when pytorch version >= 1.8.')
```

mmengine/optim/optimizer/zero_optimizer.py
```diff
@@ -0,0 +1,59 @@
...
+        assert is_available(), 'torch.distributed.rpc is not available'
```
nijkah, line 51:
```suggestion
        assert is_available(), 'torch.distributed.rpc is not available.'
```

mmengine/optim/optimizer/zero_optimizer.py
```diff
@@ -0,0 +1,59 @@
...
+        optimizer_class = getattr(torch.optim, optimizer_type)
+        super().__init__(params, optimizer_class, **kwargs)
+
+    def state_dict(self):
+        """Consolidate `state_dict`s from ranks to save the `state_dict`"""
```
nijkah, line 56:
```suggestion
        """Consolidate `state_dict`s from ranks to save the `state_dict`."""
```

mmengine/optim/optimizer/zero_optimizer.py
```diff
@@ -0,0 +1,59 @@
...
+import torch
+from torch.distributed.rpc import is_available
+
+from mmengine.dist import is_main_process
+from mmengine.utils import digit_version
+from mmengine.utils.dl_utils import TORCH_VERSION
+
+try:
+    from torch.distributed.optim import \
+        ZeroRedundancyOptimizer as _ZeroReundancyOptimizer
```
C1rN09, line 12:
```suggestion
        ZeroRedundancyOptimizer as _ZeroRedundancyOptimizer
```
Seems like a typo

nijkah:
> @nijkah Sorry for my late response. I think merely removing the `master_only` decorator is not enough, since slave processes will skip `runner.save_checkpoint` and proceed the following logics, including model removing/saving. This may lead to multiple processes attempting to remove/create a same file at the same time, which potentially results in an error. I'm not pretty sure if OS will handle this well (probably not I guess), so I think it's better to check for `is_main_process` manually and return early.

@C1rN09 , thanks for your feedback. I applied the same logic in your PR.

C1rN09 (dismissed):

zhouzaida:
Hi @nijkah , thanks for your contribution. We also need to add it in https://github.com/open-mmlab/mmengine/blob/main/docs/zh_cn/api/optim.rst and https://github.com/open-mmlab/mmengine/blob/main/docs/en/api/optim.rst.

mmengine/optim/optimizer/zero_optimizer.py
```diff
@@ -0,0 +1,59 @@
...
+            across ranks.
+        optimizer_type (str): the string of the local optimizer class.
+    """
+
+    def __init__(self, params, optimizer_type: str, **kwargs):
+        assert digit_version(TORCH_VERSION) >= digit_version('1.8.0'), (
+            '`torch.distributed.optim.ZeroReundancyOptimizer` is only '
+            'available when pytorch version >= 1.8.')
+        assert is_available(), 'torch.distributed.rpc is not available.'
+        optimizer_class = getattr(torch.optim, optimizer_type)
```
zhouzaida, line 52: Can we use `OPTIMIZERS.build` to build an optimizer? Besides, we can rename `optimizer_type`  to `optimizer` and allow it to be a dict or an instance.

mmengine/optim/optimizer/zero_optimizer.py
```diff
@@ -0,0 +1,59 @@
...
+    not divided among ranks. The partition is arbitrary and might not match the
+    the parameter registration or usage order.
+    Warnings:
+        ``ZeroRedundancyOptimizer`` requires PyTorch >= 1.8.
+    Args:
+        params (``Iterable``): an ``Iterable`` of :class:`torch.Tensor` s
+            or :class:`dict` s giving all parameters, which will be sharded
+            across ranks.
+        optimizer_type (str): the string of the local optimizer class.
+    """
```
zhouzaida, line 45:
```suggestion
    """A wrapper class of :class:`ZeroRedundancyOptimizer` that gets a
    optimizer type as string. 
    
    This class wraps an arbitrary :class:`torch.optim.Optimizer` and shards its
    states across ranks in the group as described by ZeRO_. The local optimizer
    instance in each rank is only responsible for updating approximately
    ``1 / world_size`` parameters and hence only needs to keep ``1 / world_size``
    optimizer states. After parameters are updated locally, each rank will
    broadcast its parameters to all other peers to keep all model replicas
    in the same state. ``ZeroRedundancyOptimizer`` can be used in conjunction
    with :class:`torch.nn.parallel.DistributedDataParallel` to reduce per-rank peak
    memory consumption.

    ``ZeroRedundancyOptimizer`` uses a sorted-greedy algorithm to pack a number
    of parameters at each rank. Each parameter belongs to a single rank and is
    not divided among ranks. The partition is arbitrary and might not match the
    the parameter registration or usage order.

    Warnings:
        ``ZeroRedundancyOptimizer`` requires PyTorch >= 1.8.

    Args:
        params (``Iterable``): an ``Iterable`` of :class:`torch.Tensor` s
            or :class:`dict` s giving all parameters, which will be sharded
            across ranks.
        optimizer_type (str): the string of the local optimizer class.
        
    .. _ZeRO: https://arxiv.org/abs/1910.02054
    """
```

mmengine/optim/optimizer/zero_optimizer.py
zhouzaida, line 52: Can it support custom Optimizer classes?

nijkah: I'm still figuring it out now. Until now, it does not seem to have a specific dependency on torch's optimizers. It may be possible.

C1rN09:
> Hi @C1rN09 @zhouzaida , I'm concerning about the option [`overlap_with_ddp`](https://pytorch.org/docs/master/distributed.optim.html#torch.distributed.optim.ZeroRedundancyOptimizer).
> 
> The public API docs demonstrate:
> 
> > overlap_with_ddp ([bool](https://docs.python.org/3/library/functions.html#bool), optional) – if True, [step()](https://pytorch.org/docs/master/distributed.optim.html#torch.distributed.optim.ZeroRedundancyOptimizer.step) is overlapped with DistributedDataParallel ‘s gradient synchronization; this requires (1) either a functional optimizer for the optimizer_class argument or one with a functional equivalent and (2) registering a DDP communication hook constructed from one of the functions in ddp_zero_hook.py; parameters are packed into buckets matching those in DistributedDataParallel, meaning that the parameters_as_bucket_view argument is ignored. If False, [step()](https://pytorch.org/docs/master/distributed.optim.html#torch.distributed.optim.ZeroRedundancyOptimizer.step) runs disjointly after the backward pass (per normal). (default: False)
> 
> So if someone wants to use `overlap_with_ddp` option, there should be some modifications like `registering a DDP communication hook`, `removing consolidate_state_dict`, and so on.
> 
> Will it be better to fix it as `overlap_with_ddp=False`?

Hi! If there is no easy solution, I think it's acceptable to fix `overlap_with_ddp=False` in this PR. You may add a few lines of comments/TODO to explain why. Probably we could introduce this feature in the future.

nijkah:
@C1rN09 In some versions (e.g. 1.8.0), `ZeroRedundancyOptimizer` does not have a keyword argument for `overlap_with_ddp`.
So I just added a TODO comment.

C1rN09:
Hi, @nijkah I tested your branch on my cluster and got some different results, show as below:

|  Method  |      Time       |  Memory  |
| :------: | :-------------: | :------: |
| w/o zero | ~0.24 | 7053 |
| w/  zero | ~0.24 | 6868 |

What I'm confused about is the memory consumption. Since this model is only ~250MB, it seems that the memory reduction that `ZeroOptimizer` can do with `momentum SGD` is `250MB * (1 + 1) / 4 ~ 125MB`, which is close to my results. Could you provide your detailed modifications on models, configs, etc (if any), to help me find out what I've missed?

My cluster & configurations in case I miss something important:

- Device: 4 x H100
- Configs: same as [upernet_r50_4xb4-80k_ade20k-512x512.py](https://github.com/open-mmlab/mmsegmentation/blob/1.x/configs/upernet/upernet_r50_4xb4-80k_ade20k-512x512.py), except that I changed the dataset inheritence to `cityscapes` because my cluster has no `ade20k` dataset
- Zero optimizer configs:
```python
optimizer = dict(type='ZeroRedundancyOptimizer', optimizer_type='SGD', lr=0.01, momentum=0.9, weight_decay=0.0005)
optim_wrapper = dict(type='OptimWrapper', optimizer=optimizer, clip_grad=None)
```

nijkah:
Hi, @C1rN09. I apologize for confusing this. 
You are totally correct. I found that I was confused about this today.

I was confused because I just tried to compare my result with [the provided log in github](https://download.openmmlab.com/mmsegmentation/v0.5/upernet/upernet_r50_512x512_80k_ade20k/upernet_r50_512x512_80k_ade20k_20200614_144127.log.json).

I also could get the same result as you. I'll fix the description.

C1rN09 (approved):

zhouzaida (approved):

nijkah:
nit: Hi @C1rN09. Since the SGD optimizer with momentum only stores the model params from the previous step, Should the expected memory reduction be `250MB * (1) * 0.75 ~ 175MB`, which is more close to the result?

C1rN09:
> nit: Hi @C1rN09. Since the SGD optimizer with momentum only stores the model params from the previous step, should the expected memory reduction be `250MB * (1) * 0.75 ~ 175MB`, which is more close to the result?

Yes, I think you are right! From the experiment result I guess it only shards optimizer states, instead of optimizer states + grads, which I used to think it might do.

twmht:
@nijkah 

how can i use ZeroRedundancyOptimizer with AmpOptimizer in the training config?

nijkah:
> @nijkah
> 
> how can i use ZeroRedundancyOptimizer with AmpOptimizer in the training config?

I didn't test it. Is there any specific reason to use 'AmpOptimizer` for 'mixed precision training'?
Unless it is necessary, I recommend you use `AmpOptimWrapper`. You can find a related configuration file and just change its optimizer part.

twmht:
@nijkah 

This is what i want 

```
optimizer = dict(type='ZeroRedundancyOptimizer', optimizer_type='SGD', lr=0.01, momentum=0.9, weight_decay=0.0005)
optim_wrapper = dict(type='AmpOptimWrapper', optimizer=optimizer, clip_grad=None)
```
Did you test that?

nijkah:
@twmht  Yes, it should work! 😄 
I tested it and this is a related line I edited.
https://github.com/open-mmlab/mmengine/pull/551#discussion_r1003143151

PR #555, opened 2022-09-28
[Fis] Set `_is_init` to True when `pretrained` is True

HIT-cwh:
In mmengine/hub/hub.py, the attribute `_is_init` of the model is set to True after loading checkpoint when `pretrained` is set to True so that the pretrained weights will not be overwritten.

### DIFF
--- a/mmengine/hub/hub.py
+++ b/mmengine/hub/hub.py
@@ -80,4 +80,6 @@ def get_model(cfg_path: str, pretrained: bool = False, **kwargs):
         model = MODELS.build(cfg.model, default_args=kwargs)
         if pretrained:
             load_checkpoint(model, cfg.model_path)
+            # Hack to use pretrained weights
+            model._is_init = True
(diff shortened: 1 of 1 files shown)


mmengine/hub/hub.py
```diff
@@ -80,4 +80,6 @@ def get_model(cfg_path: str, pretrained: bool = False, **kwargs):
         model = MODELS.build(cfg.model, default_args=kwargs)
         if pretrained:
             load_checkpoint(model, cfg.model_path)
+            # Hack to use pretrained weights
```
HAOCHENYE, line 83: The comments should be more detailed: "If we do not set `_is_init` here, Runner will call self.model.init_weights to overwrite the pretrained model"

HIT-cwh: Good points! Fixed

HAOCHENYE (approved):

PR #558, opened 2022-09-29
[Feature] Add `BenchmarkHook` to calculate training speed for training benchmark

ZCMax:

## Motivation

We add a benchmark hook to calculate training speed, notice that this hook should be used based on `IterTimerHook`, maybe it can be merged into `IterTimerHook`.

## Modification

add `BenchmarkHook`

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/hooks/benchmark_hook.py
+++ b/mmengine/hooks/benchmark_hook.py
@@ -1,0 +1,35 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from mmengine.registry import HOOKS
+from .hook import Hook
+
+@HOOKS.register_module()
+class BenchmarkHook(Hook):
+    """A hook that logs the training speed of each epch for benchmark."""
+
+    priority = 'NORMAL'
+
+    def after_train_epoch(self, runner) -> None:
+        """We use the average throughput in iterations of the entire training
+        run and skip the first 50 iterations of each epoch to skip GPU warmup
+        time.
+
+        Args:
+            runner (Runner): The runner of the training process.
+        """
+        message_hub = runner.message_hub
+        max_iter_num = len(runner.train_dataloader)
+        speed = message_hub.get_scalar('train/time').mean(max_iter_num - 50)
+        message_hub.update_scalar('train/speed', speed)
+        runner.logger.info(
+            f'Training speed of epoch {runner.epoch + 1} is {speed} s/iter')
+
+    def after_train(self, runner) -> None:
+        """Log average training speed of entire training process.
+
+        Args:
+            runner (Runner): The runner of the training process.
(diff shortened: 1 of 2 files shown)


mmengine/hooks/benchmark_hook.py
```diff
@@ -0,0 +1,35 @@
...
+    def after_train_epoch(self, runner) -> None:
+        """We use the average throughput in iterations of each training epcoh
+        and skip the first 50 iterations of each epoch to skip GPU warmup time.
+
+        Args:
+            runner (Runner): The runner of the training process.
+        """
+        message_hub = runner.message_hub
+        max_iter_num = len(runner.train_dataloader)
+        speed = message_hub.get_scalar('train/time').mean(max_iter_num - 50)
```
HAOCHENYE, line 21: The number "50" seems a bit magic, If `max_iter_num` is smaller than 50, it will get an unexpected result. Since we want to statistic the consumed time of an epoch, why do we still need to skip the first 50 iterations?

mmengine/hooks/benchmark_hook.py
```diff
@@ -0,0 +1,35 @@
...
+            f'Training speed of epoch {runner.epoch + 1} is {speed} s/iter')
+
+    def after_train(self, runner) -> None:
+        """Log average training speed of entire training process.
+
+        Args:
+            runner (Runner): The runner of the training process.
+        """
+        message_hub = runner.message_hub
+        avg_speed = message_hub.get_scalar('train/speed').mean()
```
HAOCHENYE, line 33: We should calculate the global average time by:
```python
message_hub.get_scalar('train/speed').mean(total_size)
```

mmengine/hooks/benchmark_hook.py
ZCMax: Generally we skip the first 50 iterations of each epoch to skip GPU warmup time. I think if `max_iter_num` is smaller than 50, we could just use the `max_iter_num` for calculation, please @ZwwWayne  have a look at this.

zhouzaida:
Please also update https://github.com/open-mmlab/mmengine/blob/main/docs/zh_cn/tutorials/hook.md.

HAOCHENYE:
> Please also update https://github.com/open-mmlab/mmengine/blob/main/docs/zh_cn/tutorials/hook.md.

Also update https://github.com/open-mmlab/mmengine/blob/main/docs/en/api/hooks.rst and https://github.com/open-mmlab/mmengine/blob/main/docs/zh_cn/api/hooks.rst

mzr1996:
What's the difference between this hook and `IterTimerHook`?

PR #556, opened 2022-09-28
[Fix] Add `ignore_keys` in ConcatDataset

BIGWangYuDong:
Sometimes there are some unnecessary keys in dataset.metainfo, but in ConcatDataset, it is forced to set all dataset.metainfo equal. In this PR, we add an `ignore_key` to avoid some unnecessary keys not equal in some concat dataset.

Related PR: https://github.com/open-mmlab/mmdetection/pull/8847

PR #559, opened 2022-09-30
[Fix] Update metainfo check in ConcatDataset

wufan-tb:
We changed the metainfo checking parts in concatdataset wrapper: from "`dataset._metainfo` must be equal" to "`dataset._metainfo['CLASSES']` must be equal".  
You see, when some one defined a dataset and set dataset._metainfo['PALETTE'] to random generation, in which case he can't concat two of this datasets just for they don't have a same 'PALETTE', we think it's really inconvenient.


ZwwWayne:
Similar issue discussed in #559, may need to find a unified solution

HAOCHENYE:
I think #556 is more flexible.

ZwwWayne:
This PR resolves https://github.com/open-mmlab/mmdetection/pull/8847

ZwwWayne:
> I think #556 is more flexible.

Great, then let's fix it in #556

PR #612, opened 2022-10-15
let runner can accept None work_dir

shenmishajing:
## Motivation

The mmengine.runner do not accept None as work_dir now, and in the downstream repositories, like mmdetection, we have to set it default value manually like this:

https://github.com/open-mmlab/mmdetection/blob/1ea191808751638f4956e51d41eace8900cc1fef/tools/train.py#L68-L75
```python
    # work_dir is determined in this priority: CLI > segment in file > filename
    if args.work_dir is not None:
        # update configs according to CLI args if args.work_dir is not None
        cfg.work_dir = args.work_dir
    elif cfg.get('work_dir', None) is None:
        # use config filename as default work_dir if cfg.work_dir is None
        cfg.work_dir = osp.join('./work_dirs',
                                osp.splitext(osp.basename(args.config))[0])
```

And, it set the work_dir to `./work_dirs/<config file name>` by default.

In mmengine.runner, we will save checkpoints, config files and modules metadata to `./work_dirs/<config file name>`, and we only save the log to `./work_dirs/<config file name>/<timestamp>`, which is dangerous. Because if we forget creating an new config file with a different name to run a new experiment, we will only got the log files for every experiments, but lost the checkpoints, which is more important. Therefore, I think we should set the work_dir to `./work_dirs/<config file name>/<timestamp>` by default, and set log_dir is equal to work_dir. And we should make this in mmengine.runner instead of let downstream repositories implement the default behavior.

## Modification

work_dir can accept None value, by default, it will be set to `./work_dirs/<config file name>/<timestamp>`.

By the way, instead of add self._timestamp to self._experiment_name, I set it to self._experiment_id, which can be used to setup vis_backends like wandb etc.

## BC-breaking (Optional)

Just delete the code for setting default value of work_dir.

change
https://github.com/open-mmlab/mmdetection/blob/1ea191808751638f4956e51d41eace8900cc1fef/tools/train.py#L68-L75
```python
    # work_dir is determined in this priority: CLI > segment in file > filename
    if args.work_dir is not None:
        # update configs according to CLI args if args.work_dir is not None
        cfg.work_dir = args.work_dir
    elif cfg.get('work_dir', None) is None:
        # use config filename as default work_dir if cfg.work_dir is None
        cfg.work_dir = osp.join('./work_dirs',
                                osp.splitext(osp.basename(args.config))[0])
```
to
```python
    # work_dir is determined in this priority: CLI > segment in file > filename
    if args.work_dir is not None:
        # update configs according to CLI args if args.work_dir is not None
        cfg.work_dir = args.work_dir
```

## Checklist

2. The documentation has been modified accordingly, like docstring or example tutorials.


zhouzaida:
Hi @shenmishajing , thanks for your contribution. In fact, if we save checkpoints to `./work_dirs/<config file name>/<timestamp>`, it is not easy to automactically resume training from `work_dir` unless we also recore the `word_dirs` in somewhere.

shenmishajing:
Yes, it's a kind of trade-off. It's no easy to automatically resume training if we automactically distinguish every experiment with same config file (name).

But, in my opinion, we run experiments with much higher frequency than automatically resume. Therefore, I think we should make the first one automatical instead of the later one, for convenience.

To resume, see #614 , we can set the experiment name and experiment id, which consistute `work_dir`, to vis_backend like wandb etc. Therefore, we can use load_from resume or spcefic work_dir resume.

How about this, we add a flag named `auto_distinguish_experiment` to runner.
If that flag is True, we add the timestamp to work_dir to automatically distinguish experiment with same config file (name), and we have to specific work_dir manually to use auto resume featuer.
In the other side, if that flag is False, we do not add timestamp, and we have to ensure that every experiment run with different config file (name), to keep the checkpoints of every experiment, and we can use the auto resume feature like now.

HAOCHENYE:
Hi, thanks for your suggestions! I highly agree that `Runner` should be able to accept `None` work_dirs. It is so annoying that I always have to struggle to come up with a `work_dir` when I write the example code :roll_eyes:.

Usually, a config file is used for an experiment of the specific Algorithm, and each `work_dirs` should only contain a single experiment directory. However, the training process could be interrupted for some reason, and we need to record these tried configs. 

If you want to record the experiment with different hyperparameters, we suggest you create a new config or modify the variable in command line like:

```bash
python tools/train.py --cfg-options <xxx=xxx> --work_dirs <new_work_dirs>
``` 

and then, your new experiment will be recorded in `new_work_dirs`.

BTW, all convention mentioned above is not strict, you can customize your experiment structure as you wish by implementing your training entry point as you like :laughing:.

shenmishajing:
> Hi, thanks for your suggestions! I highly agree that `Runner` should be able to accept `None` work_dirs. It is so annoying that I always have to struggle to come up with a `work_dir` when I write the example code 🙄.
> 
> Usually, a config file is used for an experiment of the specific Algorithm, and each `work_dirs` should only contain a single experiment directory. However, the training process could be interrupted for some reason, and we need to record these tried configs.
> 
> If you want to record the experiment with different hyperparameters, we suggest you create a new config or modify the variable in command line like:
> 
> ```shell
> python tools/train.py --cfg-options <xxx=xxx> --work_dirs <new_work_dirs>
> ```
> 
> and then, your new experiment will be recorded in `new_work_dirs`.
> 
> BTW, all convention mentioned above is not strict, you can customize your experiment structure as you wish by implementing your training entry point as you like 😆.

@HAOCHENYE I agree with you that we should create a new config to run different experiment. But, during research specially some exploration with the hyperparameters of some new modules, we may have to try so much hyperparameter combinations out. And most of them are too bad and will be thrown into trash after just a run. Therefore, it will be annoying to give an unique config (or work dir) name for every hyperparameter combinations. By the way, we may have many hyperparameters to modify in this case. I do not think write them in command line is a good idea.

As you say, it's not strict, so I think the addition of flag auto_distinguish_experiment is great, which will help user to decide whether have to propose many experiment names or not. And @zhouzaida what's your opinion?

HAOCHENYE:
> > Hi, thanks for your suggestions! I highly agree that `Runner` should be able to accept `None` work_dirs. It is so annoying that I always have to struggle to come up with a `work_dir` when I write the example code roll_eyes.
> > Usually, a config file is used for an experiment of the specific Algorithm, and each `work_dirs` should only contain a single experiment directory. However, the training process could be interrupted for some reason, and we need to record these tried configs.
> > If you want to record the experiment with different hyperparameters, we suggest you create a new config or modify the variable in command line like:
> > ```shell
> > python tools/train.py --cfg-options <xxx=xxx> --work_dirs <new_work_dirs>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then, your new experiment will be recorded in `new_work_dirs`.
> > BTW, all convention mentioned above is not strict, you can customize your experiment structure as you wish by implementing your training entry point as you like laughing.
> 
> @HAOCHENYE I agree with you that we should create a new config to run different experiment. But, during research specially some exploration with the hyperparameters of some new modules, we may have to try so much hyperparameter combinations out. And most of them are too bad and will be thrown into trash after just a run. Therefore, it will be annoying to give an unique config (or work dir) name for every hyperparameter combinations. By the way, we may have many hyperparameters to modify in this case. I do not think write them in command line is a good idea.
> 
> As you say, it's not strict, so I think the addition of flag auto_distinguish_experiment is great, which will help user to decide whether have to propose many experiment names or not. And @zhouzaida what's your opinion?

Hmmm, but if we save experiment results with different hyperparameters by different timestamps, I think it could be hard to distinguish which config belongs to which modification and the workspace will also seem messy. From my perspective, if I only want to make a small modification, I'd like to change the config in the command line, and give it a unique `work_dir` for distinguish what changes have I made. If I want to make a big change, I will create a new config inheriting from the original one, and then train with the new config(work_dir will be automatically updated.)

shenmishajing:
> 

In fact, it will not be hard to distinguish which config belongs to which modification. See pr #613 and #614 . vis_backend like wandb will automatically save the config and experiment id for you and you can compare the configurations of different experiments by it.

Wandb will save config as this.
截屏2022-10-19 12 34 24

And you can compare the config between experiments like this. It will show the different config among the experiments you selected.
截屏2022-10-19 12 36 54

HAOCHENYE:
Yeah, wandb is actually a fancy experiment tools :cowboy_hat_face:, but sometimes users could not be convenient to use wandb, and need to check the experiment results locally. So we need to take care of this case and keep the local directory structure readable and tidy.

I think a compromise solution is that we keep the current workspace structure unchanged by default(including `work_dir` and `log_dir`), and we can enable another workspace structure by passing the arguments `auto_distinguish_experiment`. This solution could also prevent from BC breaking for downstream repos.

shenmishajing:
Yes, wandb will work with the current workspace structure. But as I said, we can not get the ckpt back if we forget to change the config file name when we run a new experiment, which wandb can not help us.

Ok, I will take your solution, sounds great to me.

mmengine/runner/runner.py
```diff
@@ -236,13 +239,14 @@ class Runner:
...
         model: Union[nn.Module, Dict],
-        work_dir: str,
+        work_dir: Optional[str] = None,
         train_dataloader: Optional[Union[DataLoader, Dict]] = None,
         val_dataloader: Optional[Union[DataLoader, Dict]] = None,
         test_dataloader: Optional[Union[DataLoader, Dict]] = None,
         train_cfg: Optional[Dict] = None,
         val_cfg: Optional[Dict] = None,
         test_cfg: Optional[Dict] = None,
+        auto_distinguish_experiment: Optional[bool] = True,
```
HAOCHENYE, line 249: `auto_distinguish_experiment` could be hidden into the `cfg`.

mmengine/runner/runner.py
```diff
@@ -349,19 +350,38 @@ def __init__(
...
+
+        if work_dir is None:
+            if 'work_dir' in self.cfg and self.cfg.work_dir is not None:
+                work_dir = self.cfg.work_dir
+            elif self._experiment_name is not None:
+                if auto_distinguish_experiment:
+                    work_dir = osp.join('./work_dirs', self._experiment_name,
+                                        self._experiment_id)
+                else:
+                    work_dir = osp.join('./work_dirs', self._experiment_name)
```
HAOCHENYE, line 368: Should `auto_distinguish_experiment` be the outer `if else`, and  `work_dir`, `log_dir` is set by different branch?

mmengine/runner/runner.py
HAOCHENYE:
```suggestion
        auto_distinguish_experiment: bool = True,
```
`Optional` is not necessary (Optional is only for argument with default value `None`).

shenmishajing: > `auto_distinguish_experiment` could be hidden into the `cfg`.

Do you mean that we should delete the `auto_distinguish_experiment` param from the runner init parameters, and get it by cfg.get('auto_distinguish_experiment', True)?

shenmishajing: > `Optional` is not necessary (Optional is only for argument with default value `None`).

Got it.

mmengine/runner/runner.py
shenmishajing: emmmm……I think the current way is better, if we move auto_distinguish_experiment to outer if else, we will get so many duplicated lines.

shenmishajing: Or we can accept a little bit of logic modification as this. How about this?

HAOCHENYE:
I'm very sorry 😭, after discussion with my colleagues, the general consensus is that it's not a good idea to use two types of workspace directory structures. It is kind of confusing. So we're not going to create a different structure of `work_dir` by determining `auto_distinguish_experiment`

But I think it's acceptable for `Runner ` to accept a None `work_dir`, and we can keep this part of the PR.  
Sorry again 😞 , I should have discussed this with my colleagues before replying.

shenmishajing:
> 

Ok, and how about this, we keep all the original workspace structure, but we set the work_dir to `./work_dirs/<config file name>/<timestamp>` if `auto_distinguish_experiment` has been set?

The code will look like:
```python
        if work_dir is None:
            if 'work_dir' in self.cfg and self.cfg.work_dir is not None:
                work_dir = self.cfg.work_dir
            elif self._experiment_name is not None:
                work_dir = osp.join('./work_dirs', self._experiment_name)
            else:
                raise ValueError(
                    'work_dir, cfg.work_dir, experiment_name and cfg.filename'
                    'can not be all None.')

        self._work_dir = osp.abspath(work_dir)
        if auto_distinguish_experiment:
            self._work_dir = osp.join(self._work_dir, self._experiment_id)
        self._log_dir = osp.join(self._work_dir, self._experiment_id)
        mmengine.mkdir_or_exist(self._log_dir)
```

HAOCHENYE:
> > 
> 
> Ok, and how about this, we keep all the original workspace structure, but we set the work_dir to `./work_dirs/<config file name>/<timestamp>` if `auto_distinguish_experiment` has been set?
> 
> The code will look like:
> 
> ```python
>         if work_dir is None:
>             if 'work_dir' in self.cfg and self.cfg.work_dir is not None:
>                 work_dir = self.cfg.work_dir
>             elif self._experiment_name is not None:
>                 work_dir = osp.join('./work_dirs', self._experiment_name)
>             else:
>                 raise ValueError(
>                     'work_dir, cfg.work_dir, experiment_name and cfg.filename'
>                     'can not be all None.')
> 
>         self._work_dir = osp.abspath(work_dir)
>         if auto_distinguish_experiment:
>             self._work_dir = osp.join(self._work_dir, self._experiment_id)
>         self._log_dir = osp.join(self._work_dir, self._experiment_id)
>         mmengine.mkdir_or_exist(self._log_dir)
> ```

I think it is OK(after discussion :rofl: ). It could be better to make `auto_distinguish_experiment` defined in cfg. It means the cole wii like this:

```python
if cfg.auto_distinguish_experiment:
    self._work_dir = osp.join(self._work_dir, self._experiment_id)
```

It will not cause BC breaking and we do not need to change the signature of `Runner.from_cfg` and `Runner.__init__`.

mmengine/runner/runner.py
```diff
@@ -349,19 +346,34 @@ def __init__(
...
-        if experiment_name is not None:
-            self._experiment_name = f'{experiment_name}_{self._timestamp}'
-        elif self.cfg.filename is not None:
+        if experiment_name is None and self.cfg.filename is not None:
             filename_no_ext = osp.splitext(osp.basename(self.cfg.filename))[0]
-            self._experiment_name = f'{filename_no_ext}_{self._timestamp}'
-        else:
-            self._experiment_name = self.timestamp
-        self._log_dir = osp.join(self.work_dir, self.timestamp)
+            experiment_name = f'{filename_no_ext}'
```
HAOCHENYE, line 351: Is this line necessary? how about:
`experiment_name = osp.splitext(osp.basename(self.cfg.filename))[0]`

mmengine/runner/runner.py
```diff
@@ -349,19 +346,34 @@ def __init__(
...
+
+        self._experiment_name = experiment_name
+        self._experiment_id = self._timestamp
+
+        if work_dir is None:
+            if 'work_dir' in self.cfg and self.cfg.work_dir is not None:
```
HAOCHENYE, line 357:
```suggestion
            if self.cfg.get('word_dir') is not None:
```

mmengine/runner/runner.py
```diff
@@ -464,7 +476,8 @@ def from_cfg(cls, cfg: ConfigType) -> 'Runner':
     @property
     def experiment_name(self):
         """str: Name of experiment."""
-        return self._experiment_name
+        return f'{self._experiment_name}_{self._timestamp}' if \
+            self._experiment_name is not None else self._timestamp
```
HAOCHENYE, line 480: Why do we need to make `self.experiment_name` different from `self._self.experiment_name`

mmengine/runner/runner.py
```diff
@@ -349,19 +346,34 @@ def __init__(
...
+
+        if work_dir is None:
+            if 'work_dir' in self.cfg and self.cfg.work_dir is not None:
+                work_dir = self.cfg.work_dir
+            elif self._experiment_name is not None:
+                work_dir = osp.join('./work_dirs', self._experiment_name)
+            else:
+                raise ValueError(
+                    'work_dir, cfg.work_dir, experiment_name and cfg.filename'
+                    'can not be all None.')
```
HAOCHENYE, line 364: Should we make the f`work_dir/{self.timestamp}` as the default `work_dir`?

mmengine/runner/runner.py
```diff
@@ -349,19 +346,34 @@ def __init__(
...
+
+        self._work_dir = osp.abspath(work_dir)
+        if 'auto_distinguish_experiment' in self.cfg \
+                and self.cfg.auto_distinguish_experiment:
```
HAOCHENYE, line 368:
```suggestion
        if self.cfg.get(auto_distinguish_experiment) is True:
```

mmengine/runner/runner.py
shenmishajing: See #614 , we have to keep then name and id for vis backends.

mmengine/runner/runner.py
shenmishajing: Do you mean that we change
```python
        self._work_dir = osp.abspath(work_dir)
        if 'auto_distinguish_experiment' in self.cfg \
                and self.cfg.auto_distinguish_experiment:
            self._work_dir = osp.join(self._work_dir, self._experiment_id)
```
to
```python
        self._work_dir = osp.abspath(work_dir, , self._experiment_id)
```
and remove the auto_distinguish_experiment flag?

I can accept it, but what do you and your colleagues think about it?

mmengine/runner/runner.py
HAOCHENYE: I don't think it is an elegant way to extend the signature of `BaseVisBackend` for a specific backend. In other words, if [Neptune](https://neptune.ai/?utm_source=googleads&utm_medium=googleads&utm_campaign=brand_ads&gclid=Cj0KCQjw48OaBhDWARIsAMd966CBVg8JyQcHQSMGuBAzCwsLxTZxQSREyV0hsqottHdnsgQqoQ4LjVgaAqHbEALw_wcB) require another argument, we still need to change the interface. 

I think a workaround could be like that:

- Make `BaseVisBackend` accept **kwargs arguments.
- Set `experiment` argument in 
https://github.com/open-mmlab/mmengine/blob/dc01545e267f5bdf5afa2e1fdf781eb05f7df12c/mmengine/runner/runner.py#L391
- Parse `experiment_name` and `experiment_id` through `runner.experiment` in `WandbVisBackend`.

We may invite @zhouzaida @hhaAndroid to have look and ask for their opinion.

shenmishajing: In fact, all the Software as a Service(SaaS) Logger visbackends will take the `id` argument like wandb. See
- [Comet ML](https://www.comet.com/docs/v2/api-and-sdk/python-sdk/reference/Experiment/#:~:text=actually%20logging%20anything.-,experiment_key,-%3A%20Optional.%20If%20provided)
- [MLflow](https://mlflow.org/docs/latest/python_api/mlflow.html#mlflow.start_run)
- [Neptune](https://docs.neptune.ai/api/neptune/#init_run)
and [Wandb](https://docs.wandb.ai/ref/python/init#:~:text=this%20integration.-,id,-(str%2C%20optional)%20A).

That's why I think add the `id` param is better. I agree with you for the `**kwargs`, but as the `runner.experiment`, I think `name` and `id` will be better, because we will use them for many loggers (if we want to add more SaaS loggers). The single `runner.experiment` may cause mistake unnecessary.

But, if you persist in the `runner.experiment` way, I can go through with it.

mmengine/runner/runner.py
shenmishajing: It souds great to me.

mmengine/runner/runner.py
HAOCHENYE: Alright, I've understood. Since it is a common practice, It sounds proper to add an `experiment_name` and ``experiment`_id` argument to `BaseVisBackend`. @hhaAndroid @zhouzaida

tests/test_runner/test_runner.py
```diff
@@ -884,6 +885,32 @@ def forward(self, *args, **kwargs):
...
+        # When do not auto distinguish experiments
+        # work_dir is set to work_dirs/experiment_name/timestamp
+        # by default.
+        cfg.auto_distinguish_experiment = True
+        cfg.experiment_name = 'test_distinguish_experiment_true'
+        runner = Runner.from_cfg(cfg)
+        self.assertTrue(
+            runner.work_dir.endswith(
+                os.path.join('work_dirs', 'test_distinguish_experiment_true',
+                             runner.timestamp)))
```
HAOCHENYE, line 912: This test function can be merged into 
https://github.com/open-mmlab/mmengine/blob/601db12d38245140775ed189fcc3b4815843b567/tests/test_runner/test_runner.py#L420

HAOCHENYE: could we go to the `self.tempdir` by `os.chdir()`, and then build the runner with `None` `work_dir`？Finally we can go back to the origin directory.

shenmishajing: Got it.

tests/test_runner/test_runner.py
```diff
@@ -676,6 +677,35 @@ def test_init(self):
                                     '`param_scheduler` should be a'):
             Runner(**cfg)
 
+        # 7. Test auto distinguish experiment
+        cur_dir = os.getcwd()
+        os.chdir(self.temp_dir)
+        self.temp_dir = 'work_dirs'
```
HAOCHENYE, line 683: We should not overwrite `self.temp_dir` here, otherwise `tearDown` cannot remove the temporary dirctory.

tests/test_runner/test_runner.py
```diff
@@ -301,13 +301,14 @@ class CustomRunner(Runner):
...
                  model,
-                 work_dir,
+                 work_dir=None,
                  train_dataloader=None,
                  val_dataloader=None,
                  test_dataloader=None,
                  train_cfg=None,
                  val_cfg=None,
                  test_cfg=None,
+                 auto_distinguish_experiment=None,
```
HAOCHENYE, line 311: It seems that we do not need to modify `CustomRunner` anymore.

mmengine/runner/runner.py
```diff
@@ -349,19 +346,34 @@ def __init__(
...
 
-        if experiment_name is not None:
-            self._experiment_name = f'{experiment_name}_{self._timestamp}'
-        elif self.cfg.filename is not None:
-            filename_no_ext = osp.splitext(osp.basename(self.cfg.filename))[0]
-            self._experiment_name = f'{filename_no_ext}_{self._timestamp}'
-        else:
-            self._experiment_name = self.timestamp
-        self._log_dir = osp.join(self.work_dir, self.timestamp)
+        if experiment_name is None and self.cfg.filename is not None:
```
zhouzaida, line 349: If `cfg` does not contain `filename` key, it will throw an error.

shenmishajing: got it

PR #613, opened 2022-10-15
fix add_config func is not called bug

shenmishajing:
## Motivation

The add_config func of visualizer is not called, so all vis_backends will not get the configs.
call `self.visualizer.add_config(self.cfg)` after create visualizer.

The add_config func of wandb is wrong, see https://docs.wandb.ai/guides/track/config
wandb add_config by `wandb.config.update(<a dict object here>)` instead of `wandb.save()`

In addition, wandb can log all codes to reproduce experiments, I add this feature in add_config func.

## Modification

call `self.visualizer.add_config(self.cfg)` after create visualizer.
fix wandb add_config bug by `wandb.config.update(<a dict object here>)` instead of `wandb.save()`
add wandb log_code feature

## Use cases (Optional)

set the log_code_name of WandbVisBackend to named code artifacts.

## Checklist

1. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
2. The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/mmengine/runner/runner.py
+++ b/mmengine/runner/runner.py
@@ -389,7 +389,8 @@ class Runner:
         self.message_hub = self.build_message_hub()
         # visualizer used for writing log or visualizing all kinds of data
         self.visualizer = self.build_visualizer(visualizer)
-        self.visualizer.add_config(self.cfg)
+        if self.cfg:
+            self.visualizer.add_config(self.cfg)
 
         self._load_from = load_from
         self._resume = resume
--- a/mmengine/runner/runner.py
+++ b/mmengine/runner/runner.py
@@ -389,6 +389,7 @@ class Runner:
         self.message_hub = self.build_message_hub()
         # visualizer used for writing log or visualizing all kinds of data
         self.visualizer = self.build_visualizer(visualizer)
+        self.visualizer.add_config(self.cfg)
 
         self._load_from = load_from
         self._resume = resume
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -354,19 +354,24 @@ class WandbVisBackend(BaseVisBackend):
                 updates the current metrics dict with the row argument
                 and metrics won't be saved until `wandb.log` is called
                 with `commit=True`. Default to True.
+        log_code_name: (str, optional) The name of code artifact.
+                By default, the artifact will be named
+                source-$PROJECT_ID-$ENTRYPOINT_RELPATH. See
+                `wandb docs <https://docs.wandb.ai/ref/python/run#log_code>`_
+                for details. Default: None.
     """
 
     def __init__(self,
                  save_dir: str,
                  init_kwargs: Optional[dict] = None,
                  define_metric_cfg: Optional[dict] = None,
-                 log_code_name: Optional[str] = None,
-                 commit: Optional[bool] = True):
+                 commit: Optional[bool] = True,
+                 log_code_name: Optional[str] = None):
         super().__init__(save_dir)
         self._init_kwargs = init_kwargs
         self._define_metric_cfg = define_metric_cfg
-        self._log_code_name = log_code_name
         self._commit = commit
+        self._log_code_name = log_code_name
 
     def _init_env(self):
         """Setup env for wandb."""
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -355,10 +355,11 @@ class WandbVisBackend(BaseVisBackend):
                 and metrics won't be saved until `wandb.log` is called
                 with `commit=True`. Default to True.
         log_code_name: (str, optional) The name of code artifact.
-                By default, the artifact will be named
-                source-$PROJECT_ID-$ENTRYPOINT_RELPATH. See
-                `wandb docs <https://docs.wandb.ai/ref/python/run#log_code>`_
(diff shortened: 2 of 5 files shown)


mmengine/visualization/vis_backend.py
```diff
@@ -360,10 +360,12 @@ def __init__(self,
                  save_dir: str,
                  init_kwargs: Optional[dict] = None,
                  define_metric_cfg: Optional[dict] = None,
+                 log_code_name: Optional[str] = None,
```
zhouzaida, line 363: Adding a new paramter in middle position may cause bc issue. Suggest moving it to the last.

zhouzaida:
Hi @shenmishajing , thanks for your contribution. It would make the code more robust if unit tests are added for those modifications.

mmengine/runner/runner.py
```diff
@@ -389,6 +389,7 @@ def __init__(
         self.message_hub = self.build_message_hub()
         # visualizer used for writing log or visualizing all kinds of data
         self.visualizer = self.build_visualizer(visualizer)
+        self.visualizer.add_config(self.cfg)
```
zhouzaida, line 392:
```suggestion
        if self.cfg:
            self.visualizer.add_config(self.cfg)
```

mmengine/visualization/vis_backend.py
shenmishajing: Got it, thanks for your suggestion.

mmengine/visualization/vis_backend.py
```diff
@@ -354,17 +354,24 @@ class WandbVisBackend(BaseVisBackend):
                 updates the current metrics dict with the row argument
                 and metrics won't be saved until `wandb.log` is called
                 with `commit=True`. Default to True.
+        log_code_name: (str, optional) The name of code artifact.
+                By default, the artifact will be named
+                source-$PROJECT_ID-$ENTRYPOINT_RELPATH. See
+                `wandb docs <https://docs.wandb.ai/ref/python/run#log_code>`_
+                for details. Default: None.
```
zhouzaida, line 361:
```suggestion
        log_code_name: (str, optional) The name of code artifact.
            By default, the artifact will be named
            source-$PROJECT_ID-$ENTRYPOINT_RELPATH. See
            `wandb docs <https://docs.wandb.ai/ref/python/run#log_code>`_
            for details. Defaults to None.
            New in version 0.3.0.
```

mmengine/visualization/vis_backend.py
```diff
@@ -354,17 +354,25 @@ class WandbVisBackend(BaseVisBackend):
                 updates the current metrics dict with the row argument
                 and metrics won't be saved until `wandb.log` is called
                 with `commit=True`. Default to True.
+        log_code_name: (str, optional) The name of code artifact.
```
zhouzaida, line 357: Hi, can we rename `log_code_name` to `log_code_kwargs` which is a dict passed to `log_code` method?

shenmishajing: We can, but `log_code` method only have four param, root, name and two callable. We can only get the root and name from func kwargs. That two callables are not easy to get from vis_backend init and from config file, I think. The root is set to current path by default, any reason to change it? At least, I can not figure it out.

The only reason for this change is that we can accept two lists of regular expressions to implement the include_func and exclude_func. But, it will take many lines code to implement this, which will be a little ugrly. On the other hand, we use the python file as config file instead of yaml file, so wandb will log all config files, it seems we have no reason to change the default include and exclude func. And, that's why I only put the name param here.

What's your opinion? If you persist in the dict format, I can give it a try.

zhouzaida: The `Registry` supports registering functions so it is possible to accept two strings to provide `include_func` and `exclude_func`. Although there is no difficulty in implementation, I rather agree with your idea that there is no need to support full features of `log_code` currently.

shenmishajing: Ok, I got it, but can you explain more about it? Is there any docs or code examples? I think I need some help to implement this.

shenmishajing:
@zhouzaida How about this? I add a registry name PATH_FILTERS to specific the include_fn and exclude_fn.

zhouzaida:
Sorry, I may have misrepresented it, what I was trying to say was that you could follow your previous solution and not support include_func and exclude_func.

shenmishajing:
> Sorry, I may have misrepresented it, what I was trying to say was that you could follow your previous solution and not support include_func and exclude_func.

@zhouzaida Oh……I'm sorry I ignore the `no need`'s `no` in your comment. So, should I roll it back?

zhouzaida:
> > Sorry, I may have misrepresented it, what I was trying to say was that you could follow your previous solution and not support include_func and exclude_func.
> 
> @zhouzaida Oh……I'm sorry I ignore the `no need`'s `no` in your comment. So, should I roll it back?

Suggest rolling it back.

shenmishajing:
> > > Sorry, I may have misrepresented it, what I was trying to say was that you could follow your previous solution and not support include_func and exclude_func.
> > 
> > 
> > @zhouzaida Oh……I'm sorry I ignore the `no need`'s `no` in your comment. So, should I roll it back?
> 
> Suggest rolling it back.

@zhouzaida Got it.

zhouzaida (approved):

PR #617, opened 2022-10-16
support set omp and mkl num threads in set_multi_processing func

shenmishajing:
## Motivation

Support set omp and mkl num threads in set_multi_processing func. Therefore, users can set them in mp_cfg.

## Modification

Add omp_num_threads and mkl_num_threads params to set_multi_processing func.

## Use cases (Optional)

Now, we can set omp and mkl num threads like:
```python
    mp_cfg = dict(
        mp_start_method='spawn',
        opencv_num_threads=4,
        omp_num_threads=5,
        mkl_num_threads=6,
        distributed=True)
```

## Checklist

### DIFF
--- a/mmengine/utils/dl_utils/setup_env.py
+++ b/mmengine/utils/dl_utils/setup_env.py
@@ -2,12 +2,15 @@
 import os
 import platform
 import warnings
+from typing import Optional
 
 import torch.multiprocessing as mp
 
 
 def set_multi_processing(mp_start_method: str = 'fork',
                          opencv_num_threads: int = 0,
+                         omp_num_threads: Optional[int] = None,
+                         mkl_num_threads: Optional[int] = None,
                          distributed: bool = False) -> None:
     """Set multi-processing related environment.
 
@@ -16,6 +19,10 @@ def set_multi_processing(mp_start_method: str = 'fork',
             child processes. Defaults to 'fork'.
         opencv_num_threads (int): Number of threads for opencv.
             Defaults to 0.
+        omp_num_threads (int, optional): Number of threads for OMP.
+            Defaults to 1 when distributed is True, otherwise no change.
+        mkl_num_threads (int, optional): Number of threads for MKL.
+            Defaults to 1 when distributed is True, otherwise no change.
         distributed (bool): True if distributed environment.
             Defaults to False.
     """
@@ -39,23 +46,26 @@ def set_multi_processing(mp_start_method: str = 'fork',
     except ImportError:
         pass
 
-    # setup OMP threads
-    # This code is referred from https://github.com/pytorch/pytorch/blob/master/torch/distributed/run.py  # noqa
-    if 'OMP_NUM_THREADS' not in os.environ and distributed:
-        omp_num_threads = 1
-        warnings.warn(
-            'Setting OMP_NUM_THREADS environment variable for each process'
-            f' to be {omp_num_threads} in default, to avoid your system '
-            'being overloaded, please further tune the variable for '
-            'optimal performance in your application as needed.')
-        os.environ['OMP_NUM_THREADS'] = str(omp_num_threads)
+    if distributed:
+        # setup OMP threads
+        # This code is referred from https://github.com/pytorch/pytorch/blob/master/torch/distributed/run.py  # noqa
+        if omp_num_threads is None and 'OMP_NUM_THREADS' not in os.environ:
+            omp_num_threads = 1
+            warnings.warn(
+                'Setting OMP_NUM_THREADS environment variable for each process'
+                f' to be {omp_num_threads} in default, to avoid your system '
+                'being overloaded, please further tune the variable for '
+                'optimal performance in your application as needed.')
+        if omp_num_threads is not None:
+            os.environ['OMP_NUM_THREADS'] = str(omp_num_threads)
 
-    # setup MKL threads
-    if 'MKL_NUM_THREADS' not in os.environ and distributed:
-        mkl_num_threads = 1
-        warnings.warn(
-            'Setting MKL_NUM_THREADS environment variable for each process'
-            f' to be {mkl_num_threads} in default, to avoid your system '
-            'being overloaded, please further tune the variable for '
-            'optimal performance in your application as needed.')
-        os.environ['MKL_NUM_THREADS'] = str(mkl_num_threads)
+        # setup MKL threads
+        if mkl_num_threads is None and 'MKL_NUM_THREADS' not in os.environ:
+            mkl_num_threads = 1
+            warnings.warn(
+                'Setting MKL_NUM_THREADS environment variable for each process'
+                f' to be {mkl_num_threads} in default, to avoid your system '
+                'being overloaded, please further tune the variable for '
+                'optimal performance in your application as needed.')
+        if mkl_num_threads is not None:
+            os.environ['MKL_NUM_THREADS'] = str(mkl_num_threads)
--- a/mmengine/utils/dl_utils/setup_env.py
+++ b/mmengine/utils/dl_utils/setup_env.py
@@ -46,26 +46,27 @@ def set_multi_processing(mp_start_method: str = 'fork',
     except ImportError:
         pass
 
-    if distributed:
-        # setup OMP threads
-        # This code is referred from https://github.com/pytorch/pytorch/blob/master/torch/distributed/run.py  # noqa
-        if omp_num_threads is None and 'OMP_NUM_THREADS' not in os.environ:
-            omp_num_threads = 1
+    # setup OMP threads
+    # This code is referred from https://github.com/pytorch/pytorch/blob/master/torch/distributed/run.py  # noqa
+    if omp_num_threads is None:
+        if 'OMP_NUM_THREADS' not in os.environ and distributed:
+            os.environ['OMP_NUM_THREADS'] = '1'
             warnings.warn(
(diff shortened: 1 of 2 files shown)


mmengine/utils/dl_utils/setup_env.py
```diff
@@ -2,12 +2,15 @@
...
 import warnings
+from typing import Optional
 
 import torch.multiprocessing as mp
 
 
 def set_multi_processing(mp_start_method: str = 'fork',
                          opencv_num_threads: int = 0,
+                         omp_num_threads: Optional[int] = None,
+                         mkl_num_threads: Optional[int] = None,
```
zhouzaida, line 13: Adding a new parameter in the middle position may cause a bc issue. Suggest moving it to the last.

shenmishajing: Got it

mmengine/utils/dl_utils/setup_env.py
```diff
@@ -39,23 +46,26 @@ def set_multi_processing(mp_start_method: str = 'fork',
...
+        # This code is referred from https://github.com/pytorch/pytorch/blob/master/torch/distributed/run.py  # noqa
+        if omp_num_threads is None and 'OMP_NUM_THREADS' not in os.environ:
+            omp_num_threads = 1
+            warnings.warn(
+                'Setting OMP_NUM_THREADS environment variable for each process'
+                f' to be {omp_num_threads} in default, to avoid your system '
+                'being overloaded, please further tune the variable for '
+                'optimal performance in your application as needed.')
+        if omp_num_threads is not None:
+            os.environ['OMP_NUM_THREADS'] = str(omp_num_threads)
```
HAOCHENYE, line 60:
```suggestion
    if not distributed:
        return 

    if omp_num_threads is not None:
        os.environ['OMP_NUM_THREADS'] = str(omp_num_threads)
    else:
        omp_num_threads = 1
        warnings.warn(
            'Setting OMP_NUM_THREADS environment variable for each process'
            f' to be {omp_num_threads} in default, to avoid your system '
            'being overloaded, please further tune the variable for '
            'optimal performance in your application as needed.')
    
```

shenmishajing: Should we set OMP_NUM_THREADS and KML_NUM_THREADS when not distributed?
How about change whole thing to:
```python
    # setup OMP threads
    # This code is referred from https://github.com/pytorch/pytorch/blob/master/torch/distributed/run.py  # noqa
    if omp_num_threads is None and 'OMP_NUM_THREADS' not in os.environ and distributed:
        omp_num_threads = 1
        warnings.warn(
            'Setting OMP_NUM_THREADS environment variable for each process'
            f' to be {omp_num_threads} in default, to avoid your system '
            'being overloaded, please further tune the variable for '
            'optimal performance in your application as needed.')
    if omp_num_threads is not None:
        os.environ['OMP_NUM_THREADS'] = str(omp_num_threads)

    # setup MKL threads
    if mkl_num_threads is None and 'MKL_NUM_THREADS' not in os.environ and distributed:
        mkl_num_threads = 1
        warnings.warn(
            'Setting MKL_NUM_THREADS environment variable for each process'
            f' to be {mkl_num_threads} in default, to avoid your system '
            'being overloaded, please further tune the variable for '
            'optimal performance in your application as needed.')
    if mkl_num_threads is not None:
        os.environ['MKL_NUM_THREADS'] = str(mkl_num_threads)
```

HAOCHENYE: It seems `omp_num_threads` and  `mkl_num_threads` have the highest priority, how about check `omp_num_threads` is None or not firstly :joy:?  Maybe it could be more readable. In the current implementation, `omp_num_threads` could be checked twice, it seems a little bit weird.

shenmishajing: Do you mean this?
```python
    if omp_num_threads is None:
        if 'OMP_NUM_THREADS' not in os.environ and distributed:
            os.environ['OMP_NUM_THREADS'] = '1'
            warnings.warn(
                'Setting OMP_NUM_THREADS environment variable for each process'
                ' to be 1 in default, to avoid your system being overloaded, '
                'please further tune the variable for optimal performance in '
                'your application as needed.')
    else:
        os.environ['OMP_NUM_THREADS'] = str(omp_num_threads)

```

HAOCHENYE: > Do you mean this?
> 
> ```python
>     if omp_num_threads is None:
>         if 'OMP_NUM_THREADS' not in os.environ and distributed:
>             os.environ['OMP_NUM_THREADS'] = '1'
>             warnings.warn(
>                 'Setting OMP_NUM_THREADS environment variable for each process'
>                 ' to be 1 in default, to avoid your system being overloaded, '
>                 'please further tune the variable for optimal performance in '
>                 'your application as needed.')
>     else:
>         os.environ['OMP_NUM_THREADS'] = str(omp_num_threads)
> ```

LGTM, Is it necessary to set `omp_num_threads` when not distributed?

HAOCHENYE: It is hard to say if we do not set `OMP_NUM_THREADS`, the default value will be 1 for 3rd party libraries. 

image

The main reason we set `OMP_NUM_THREADS=1` here is that if each process enable too many threads during ddp training, system could be overloaded.

shenmishajing: Well, I'm not very familiar with this (and that's why I leave the code here 😂). I know the origin code here is to avoid overload. But, I add the two param here because I want to get speed up from the multi threads😂. I think we should set OMP_NUM_THREADS even when not distributed, which may bring a considerable speed-up.

HAOCHENYE: Alright, I think it is also reasonable. Maybe we can invite @zhouzaida to have a look.

RangiLyu (changes requested):
Thanks for your contribution! 
The purpose of this feature is to avoid overloading during distributed training when not setting OMP and MKL environment variables. So MMEngine should not change the value when non-distributed training.
Therefore, the values set in the environment variable should be the highest priority, and MMEngine will only control the default values when these environment variables are not set during distributed training.

How about renaming these args to `default_dist_omp_threads` and `default_dist_mkl_threads` and they only work when these environment variables are not set during distributed training?

HAOCHENYE:
> Thanks for your contribution! The purpose of this feature is to avoid overloading during distributed training when not setting OMP and MKL environment variables. So MMEngine should not change the value when non-distributed training. Therefore, the values set in the environment variable should be the highest priority, and MMEngine will only control the default values when these environment variables are not set during distributed training.
> 
> How about renaming these args to `default_dist_omp_threads` and `default_dist_mkl_threads` and they only work when these environment variables are not set during distributed training?

Suppose I use non-DDP training and the bottleneck of training speed is the data transforming in the pipeline. Considering that Opencv may do parallel acceleration based on OpenMP. I may need to modify `OMP_NUM_THREADS` to free up more computational resources of the CPU. I'm not sure is this case reasonable to modify `OMP_NUM_THREADS` during non-DDP training.

RangiLyu:
> > Thanks for your contribution! The purpose of this feature is to avoid overloading during distributed training when not setting OMP and MKL environment variables. So MMEngine should not change the value when non-distributed training. Therefore, the values set in the environment variable should be the highest priority, and MMEngine will only control the default values when these environment variables are not set during distributed training.
> > How about renaming these args to `default_dist_omp_threads` and `default_dist_mkl_threads` and they only work when these environment variables are not set during distributed training?
> 
> Suppose I use non-DDP training and the bottleneck of training speed is the data transforming in the pipeline. Considering that Opencv may do parallel acceleration based on OpenMP. I may need to modify `OMP_NUM_THREADS` to free up more computational resources of the CPU. I'm not sure is this case reasonable to modify `OMP_NUM_THREADS` during non-DDP training.

I think the proper way is to set these environment variables directly(e.g. run `export OMP_NUM_THREADS=1`) instead of modifying them through mmengine's configuration.

shenmishajing:
@HAOCHENYE @RangiLyu So, how is it going? Which way should we choose? Will we set the num_threads when not distributed?

HAOCHENYE:
@shenmishajing Hi, sorry for my late reply, after discussion. After discussion, a common consensus is that we should not support configuring the arguments which could also be configured by the system environment. For example, the available GPU could also be configured by `CUDA_VISIBLE_DEVICES`, but we'll not set in training entry point like `os.environ['CUDA_VISIBLE_DEVICES'] = xxx`. We suggest user to config it in the terminal: `export CUDA_VISIBLE_DEVICES=xxx`. There is so many environment variables that can be configured, and it is not proper to configure them through the runner.

shenmishajing:
> @shenmishajing Hi, sorry for my late reply. After discussion, a common consensus is that we should not support configuring the arguments which could also be configured by the system environment. For example, the available GPU could also be configured by `CUDA_VISIBLE_DEVICES`, but we'll not set in training entry point like `os.environ['CUDA_VISIBLE_DEVICES'] = xxx`. We suggest user to config it in the terminal: `export CUDA_VISIBLE_DEVICES=xxx`. There is so many environment variables that can be configured, and it is not proper to configure them through the runner.

@HAOCHENYE Ok, got it. And, should I change this to do not set the two env vars when not disturbited or just close this pr?

HAOCHENYE:
It is ok to close this PR, and thanks for your contributions anyway :smile: !

shenmishajing:
Got it.

PR #628, opened 2022-10-18
[Enhance] add param to customized Formatter

zhijiejia:
## Motivation

If the output information is too long, the output information will occupy two lines, which seems to be very uncomfortable to user, so can we add a parameter to allow users to customize logging.Formatter

## Modification

 add param in `mmengine/logging/logger.py` for class `MMLogger`

## Checklist

### DIFF
--- a/mmengine/logging/logger.py
+++ b/mmengine/logging/logger.py
@@ -136,11 +136,11 @@ class MMLogger(Logger, ManagerMixin):
             ``FileHandler`` will be added to the logger. Defaults to None.
         log_level (str): The log level of the handler and logger. Defaults to
             "NOTSET".
-        log_datefmt (str): The log date formatter. If `log_datefmt` is not
-            defined, defaults to '%m/%d %H:%M:%S'.
         file_mode (str): The file mode used to open log file. Defaults to 'w'.
         distributed (bool): Whether to save distributed logs, Defaults to
             false.
+        log_datefmt (str): The log date formatter. If `log_datefmt` is not
+            defined, defaults to '%m/%d %H:%M:%S'.
     """
 
     def __init__(self,
@@ -148,9 +148,9 @@ class MMLogger(Logger, ManagerMixin):
                  logger_name='mmengine',
                  log_file: Optional[str] = None,
                  log_level: str = 'INFO',
-                 log_datefmt: str = '%m/%d %H:%M:%S',
                  file_mode: str = 'w',
-                 distributed=False):
+                 distributed=False,
+                 log_datefmt: str = '%m/%d %H:%M:%S'):
         Logger.__init__(self, logger_name)
         ManagerMixin.__init__(self, name)
         # Get rank in DDP mode.
--- a/mmengine/logging/logger.py
+++ b/mmengine/logging/logger.py
@@ -136,6 +136,8 @@ class MMLogger(Logger, ManagerMixin):
             ``FileHandler`` will be added to the logger. Defaults to None.
         log_level (str): The log level of the handler and logger. Defaults to
             "NOTSET".
+        log_datefmt (str): The log date formatter. If `log_datefmt` is not
+            defined, defaults to '%m/%d %H:%M:%S'.
         file_mode (str): The file mode used to open log file. Defaults to 'w'.
         distributed (bool): Whether to save distributed logs, Defaults to
             false.
@@ -146,6 +148,7 @@ class MMLogger(Logger, ManagerMixin):
                  logger_name='mmengine',
                  log_file: Optional[str] = None,
                  log_level: str = 'INFO',
+                 log_datefmt: str = '%m/%d %H:%M:%S',
                  file_mode: str = 'w',
                  distributed=False):
         Logger.__init__(self, logger_name)
@@ -160,7 +163,7 @@ class MMLogger(Logger, ManagerMixin):
         # `StreamHandler` record month, day, hour, minute, and second
         # timestamp.
         stream_handler.setFormatter(
-            MMFormatter(color=True, datefmt='%m/%d %H:%M:%S'))
+            MMFormatter(color=True, datefmt=log_datefmt))
         # Only rank0 `StreamHandler` will log messages below error level.
         stream_handler.setLevel(log_level) if rank == 0 else \
             stream_handler.setLevel(logging.ERROR)
--- a/mmengine/logging/logger.py
+++ b/mmengine/logging/logger.py
@@ -139,7 +139,7 @@ class MMLogger(Logger, ManagerMixin):
         file_mode (str): The file mode used to open log file. Defaults to 'w'.
         distributed (bool): Whether to save distributed logs, Defaults to
             false.
-        log_datefmt (str): The log date formatter. If `log_datefmt` is not
+        datefmt (str): The log date formatter. If `datefmt` is not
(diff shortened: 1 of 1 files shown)


mmengine/logging/logger.py
```diff
@@ -146,6 +148,7 @@ def __init__(self,
                  logger_name='mmengine',
                  log_file: Optional[str] = None,
                  log_level: str = 'INFO',
+                 log_datefmt: str = '%m/%d %H:%M:%S',
```
HAOCHENYE, line 151: To avoid BC breaking, we should set `log_datefmt` as the last argument.

HAOCHENYE (commented):
Thanks for your contribution! 

Currently, `Runner.__init__` does not support customize the logger because the logger is built as following:

https://github.com/open-mmlab/mmengine/blob/09a195b24cfff85d2de29f0786bbcc9f0586002c/mmengine/runner/runner.py#L370

`build_loggger` only accepts `log_level` now, and we cannot customize the logger by a simple string.

One smooth way to solve this problem is to change the function signature of `Runner.__init__`. We should rename `log_level` to `log_cfg`, which could be a `dict`. To avoid BC breaking, we should also decorate `Runner.__init__` with 
https://github.com/open-mmlab/mmengine/blob/09a195b24cfff85d2de29f0786bbcc9f0586002c/mmengine/utils/misc.py#L299, which will map the old arguments `log_level` to `log_cfg`.  We should also make sure that `log_cfg` could be a string, the old code should also build runner successfully by passing a string `log_level`.
 
Finnaly, we can deprecate the `log_level` parameter in the future safely (Maybe MMEngine 1.0.0 :rofl: )

mmengine/logging/logger.py
zhijiejia: Yes, that is better. BTW, does the `BC-Breaking` means modifications in this repo should not affect the use of downstream repo which build on this repo?

HAOCHENYE: Yes, Since downstream repo like MMDet, MMCls .etc. have refactored the code based on MMEngine, we should ensure the modification would not break their code immediately.

HAOCHENYE:
> > Thanks for your contribution!
> > Currently, `Runner.__init__` does not support customize the logger because the logger is built as following:
> > https://github.com/open-mmlab/mmengine/blob/09a195b24cfff85d2de29f0786bbcc9f0586002c/mmengine/runner/runner.py#L370
> > 
> > `build_loggger` only accepts `log_level` now, and we cannot customize the logger by a simple string.
> > One smooth way to solve this problem is to change the function signature of `Runner.__init__`. We should rename `log_level` to `log_cfg`, which could be a `dict`. To avoid BC breaking, we should also decorate `Runner.__init__` with
> > https://github.com/open-mmlab/mmengine/blob/09a195b24cfff85d2de29f0786bbcc9f0586002c/mmengine/utils/misc.py#L299
> > 
> > , which will map the old arguments `log_level` to `log_cfg`. We should also make sure that `log_cfg` could be a string, the old code should also build runner successfully by passing a string `log_level`.
> > Finnaly, we can deprecate the `log_level` parameter in the future safely (Maybe MMEngine 1.0.0 rofl )
> 
> Yes, I understand the code logic you said, but do you mean you can't add this PR now? Do you need me to change what you said?



> > Thanks for your contribution!
> > Currently, `Runner.__init__` does not support customize the logger because the logger is built as following:
> > https://github.com/open-mmlab/mmengine/blob/09a195b24cfff85d2de29f0786bbcc9f0586002c/mmengine/runner/runner.py#L370
> > 
> > `build_loggger` only accepts `log_level` now, and we cannot customize the logger by a simple string.
> > One smooth way to solve this problem is to change the function signature of `Runner.__init__`. We should rename `log_level` to `log_cfg`, which could be a `dict`. To avoid BC breaking, we should also decorate `Runner.__init__` with
> > https://github.com/open-mmlab/mmengine/blob/09a195b24cfff85d2de29f0786bbcc9f0586002c/mmengine/utils/misc.py#L299
> > 
> > , which will map the old arguments `log_level` to `log_cfg`. We should also make sure that `log_cfg` could be a string, the old code should also build runner successfully by passing a string `log_level`.
> > Finnaly, we can deprecate the `log_level` parameter in the future safely (Maybe MMEngine 1.0.0 rofl )
> 
> Yes, I understand the code logic you said, but do you mean you can't add this PR now? Do you need me to change what you said?

We're committed to building an active community to build MMEngine, so we're veeeeerry looking forward to working with you on this feature :rocket:. Of course, modifying `MMLogger` is a good start, if you're not convenient to make further changes, we'll support this feature ASAP and @ you if we've finished it.

zhijiejia:
> > > Thanks for your contribution!
> > > Currently, `Runner.__init__` does not support customize the logger because the logger is built as following:
> > > https://github.com/open-mmlab/mmengine/blob/09a195b24cfff85d2de29f0786bbcc9f0586002c/mmengine/runner/runner.py#L370
> > > 
> > > `build_loggger` only accepts `log_level` now, and we cannot customize the logger by a simple string.
> > > One smooth way to solve this problem is to change the function signature of `Runner.__init__`. We should rename `log_level` to `log_cfg`, which could be a `dict`. To avoid BC breaking, we should also decorate `Runner.__init__` with
> > > https://github.com/open-mmlab/mmengine/blob/09a195b24cfff85d2de29f0786bbcc9f0586002c/mmengine/utils/misc.py#L299
> > > 
> > > , which will map the old arguments `log_level` to `log_cfg`. We should also make sure that `log_cfg` could be a string, the old code should also build runner successfully by passing a string `log_level`.
> > > Finnaly, we can deprecate the `log_level` parameter in the future safely (Maybe MMEngine 1.0.0 rofl )
> > 
> > 
> > Yes, I understand the code logic you said, but do you mean you can't add this PR now? Do you need me to change what you said?
> 
> > > Thanks for your contribution!
> > > Currently, `Runner.__init__` does not support customize the logger because the logger is built as following:
> > > https://github.com/open-mmlab/mmengine/blob/09a195b24cfff85d2de29f0786bbcc9f0586002c/mmengine/runner/runner.py#L370
> > > 
> > > `build_loggger` only accepts `log_level` now, and we cannot customize the logger by a simple string.
> > > One smooth way to solve this problem is to change the function signature of `Runner.__init__`. We should rename `log_level` to `log_cfg`, which could be a `dict`. To avoid BC breaking, we should also decorate `Runner.__init__` with
> > > https://github.com/open-mmlab/mmengine/blob/09a195b24cfff85d2de29f0786bbcc9f0586002c/mmengine/utils/misc.py#L299
> > > 
> > > , which will map the old arguments `log_level` to `log_cfg`. We should also make sure that `log_cfg` could be a string, the old code should also build runner successfully by passing a string `log_level`.
> > > Finnaly, we can deprecate the `log_level` parameter in the future safely (Maybe MMEngine 1.0.0 rofl )
> > 
> > 
> > Yes, I understand the code logic you said, but do you mean you can't add this PR now? Do you need me to change what you said?
> 
> We're committed to building an active community to build MMEngine, so we're veeeeerry looking forward to working with you on this feature 🚀. Of course, modifying `MMLogger` is a good start, if you're not convenient to make further changes, we'll support this feature ASAP and @ you if we've finished it.


Ok, thanks, I'm still a newbie. For submitting a PR, I'll wait for you to modify it first, and I hope I can complete my first PR in ｀mmengine｀.

PR #630, opened 2022-10-18
[Fix] Fix offline eval dataset index error.

RangiLyu:

## Motivation

BaseDataset does not support slice index which is used in `offline_evaluate`.

## Modification

Replace slice index with `get_chunks` to get a batch of data.

## BC-breaking (Optional)

None.

### DIFF
--- a/mmengine/evaluator/evaluator.py
+++ b/mmengine/evaluator/evaluator.py
@@ -104,12 +104,6 @@ class Evaluator:
         """
 
         # support chunking iterable objects
-        if data is not None:
-            assert len(data_samples) == len(data), (
-                'outputs and data should have the same length, but got '
-                f'outputs length: {len(data_samples)} '
-                f'data length: {len(data)}')
-
         def get_chunks(seq: Iterator, chunk_size=1):
             stop = False
             while not stop:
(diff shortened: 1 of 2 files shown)


mmengine/evaluator/evaluator.py
```diff
@@ -123,10 +117,17 @@ def get_chunks(seq: Iterator, chunk_size=1):
                 if chunk:
                     yield chunk
 
+        if data is not None:
+            assert len(data_samples) == len(data), (
+                'outputs and data should have the same length, but got '
+                f'outputs length: {len(data_samples)} '
+                f'data length: {len(data)}')
```
HAOCHENYE, line 124:
```suggestion
                'data_samples and data should have the same length, but got '
                f'data_samples length: {len(data_samples)} '
                f'data length: {len(data)}')
```

mmengine/evaluator/evaluator.py
```diff
@@ -123,10 +117,17 @@ def get_chunks(seq: Iterator, chunk_size=1):
...
+            data = get_chunks(iter(data), chunk_size)
+
         size = 0
         for output_chunk in get_chunks(iter(data_samples), chunk_size):
             if data is not None:
-                data_chunk = pseudo_collate(data[size:size + chunk_size])
+                data_chunk = pseudo_collate(next(data))  # type: ignore
```
HAOCHENYE, line 130: should the data here be chunked?

```suggestion
                data_chunk = pseudo_collate(get_chunks(iter(data), chunk_size))  # type: ignore
```

RangiLyu: the `data` here is a chunk generator

HAOCHENYE (approved):

PR #631, opened 2022-10-18
[Enhance] Add runner test case.

HAOCHENYE:

## Motivation

All unit tests need to build runner could inherit `TestRunnerTestCase`.

## Modification

1. A lot of unit test needs to build a `Runner` instance, `RunnerTestCase` provides a simple way to build runner with iter based or epoch-based config.
2. Building a `Runner` will leave a lot of global variable, `RunnerTestCase` will clean then after each unit test.
3. `RunnerTestCase` will deinitialize the distributed enviroment.
4. `Registry` instance like `MODELS` and `HOOKS` are also global variables,  `RunnerTestCase` will also clean them after each unit test,

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

5. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
6. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
7. The documentation has been modified accordingly, like docstring or example tutorials.


mmengine/testing/runner_test_case.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def process(self, data_batch, predictions):
+        result = {'acc': 1}
+        self.results.append(result)
+
+    def compute_metrics(self, results):
+        return dict(acc=1)
+
+
+class RunnerTestCase(TestCase):
```
C1rN09, line 82: I think it's better to add some docstring here, which may help understand what it's designed for and how to use it.

mmengine/testing/runner_test_case.py
```diff
@@ -0,0 +1,172 @@
...
+        self.epoch_based_cfg = Config(epoch_based_cfg)
+        self.iter_based_cfg: Config = copy.deepcopy(self.epoch_based_cfg)
+        self.iter_based_cfg.train_dataloader = dict(
+            dataset=dict(type='ToyDataset'),
+            sampler=dict(type='InfiniteSampler', shuffle=True),
+            batch_size=3,
+            num_workers=0)
+        self.iter_based_cfg.log_processor = dict(by_epoch=False)
+
+        # prepare iter based cfg.
```
zhouzaida, line 134:
```suggestion

        # prepare iter based cfg
        self.iter_based_cfg: Config = copy.deepcopy(self.epoch_based_cfg)
        self.iter_based_cfg.train_dataloader = dict(
            dataset=dict(type='ToyDataset'),
            sampler=dict(type='InfiniteSampler', shuffle=True),
            batch_size=3,
            num_workers=0)
        self.iter_based_cfg.log_processor = dict(by_epoch=False)
```

tests/test_runner/test_runner.py
```diff
@@ -1487,7 +1511,7 @@ def before_val_iter(self, runner, batch_idx, data_batch=None):
         val_interval_results = []
         val_interval_targets = [5] * 10 + [2] * 2
 
-        @HOOKS.register_module()
+        @HOOKS.register_module(force=True)
```
zhouzaida, line 1514: Why do we need to set `force=True`?

tests/test_model/test_model_utils.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, model):
 
         pass
 
-    CHILD_REGISTRY.register_module(module=CustomModelWrapper)
+    CHILD_REGISTRY.register_module(module=CustomModelWrapper, force=True)
```
zhouzaida, line 82: Same as above

mmengine/testing/runner_test_case.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def build_runner(self, cfg):
+        cfg.experiment_name = self.experiment_name
+        runner = Runner.from_cfg(cfg)
+        return runner
+
+    @property
+    def experiment_name(self):
+        return f'{self._testMethodName}_{time.time()} + ' \
+               f'{uuid4()}'
```
zhouzaida, line 164: Add a comment to describe why we need to append `uuid4()`.

tests/test_runner/test_runner.py
HAOCHENYE: Some toy classes with the same name are defined in different unit tests. If `force=False`, the class defined in current unit test will not be registered. If we refactor the unit test and make sure each unit test will clean the registered class, there is no need to set `force=True` anymore.

mmengine/testing/runner_test_case.py
```diff
@@ -0,0 +1,186 @@
...
+    def compute_metrics(self, results):
+        return dict(acc=1)
+
+
+class RunnerTestCase(TestCase):
+    """A test case to build runner easily.
+
+    `RunnerTestCase` will do the following things:
+
+    1. Registers a toy model, toy metric, toy dataset, which can be used to
```
zhouzaida, line 87:
```suggestion
    1. Registers a toy model, a toy metric, and a toy dataset, which can be used to
```

zhouzaida (approved):

ZwwWayne (approved):

Issue #621, opened 2022-10-17
How does the add_graph func work?


shenmishajing:
The func add_graph of vis_backend is not called now. It has two parameters model and data_batch, which may indicate that it should be called every iter? But vis_backend like wandb will draw the model without data_batch and should be called only once. Therefore, should we delete the data_batch input of add_graph func? If yes, I will create a pr to do this and fix the add_graph func is not called bug.

HAOCHENYE:
Thannnnks for your feedback！ We want to draw the model in tensorboard initially, and tensorboard require `data_batch` and `model`  arguments. Since the visualization result in tensorboard is unattractive and more restrictive, we've not integrated `add_graph` in hooks now(implemented in `before_run` is acceptable).

I'm not familiar with wandb, can wandb plot the model graph without tracing the model(without data_batch)?

shenmishajing:
Yes, wandb can do this by wandb.watch. it works without `data_batch`, by hook. But, if you mean plot a png file which pictures the relationship between modules of your model by "plot the model", then the answer is no, hhhhh.

I try to add this feature into mmengine, but the log graph does not work now, so I find some result on web. See here https://wandb.ai/ayush-thakur/debug-neural-nets/runs/jh061uaf/model wandb will get the input and output of every modules in your model, which is unattractive too, I think.

In fact, the useful point of wandb.watch is it can log the disturibution of gradients and parameters of your model during training, which is usefull if your want to know is any gradient flow bottleneck in your model and will help you to redesign your network architecture.

Here is a log result(an experiment I just run about 2000 iter, gradients and parameters do not change a lot).
<img width="1804" alt="截屏2022-10-17 19 33 02" src="https://user-images.githubusercontent.com/32559715/196166746-1c9977ba-c7e5-402e-b6de-7940d87ebd53.png">

HAOCHENYE:
Wow! the weight&gradient distribution displayed by wandb seems so cool 😎! I think this feature should be integrated with `WandbVisBackend.add_graph`, and called by `NativeVisualizationHook.before_train` (model weight distribution graph and model topology graph are both graph 🤣). 

We should still keep the `data_batch` arguments since other backends need the input tensor to trace the whole model (Although `wandb.watch` utilize the backward_fn to reconstruct the graph, but utilizing the input tensor maybe could trace the graph prettier 🤔 ) .

[...]

PR #632, opened 2022-10-19
fix add graph func is not called bug

shenmishajing:
## Motivation

Fix add graph func is not called bug. See issue #621 

## Modification

- Call visualizer.add_graph after model build.
- Add watch_kwargs param to WandbVisBackend.

## Checklist


HAOCHENYE:
Will it be more proper to call `add_graph` in `NativeVisualizerHook.before_train`, and this feature can be enabled by setting
`watch_model=True` (default False, not sure will it harm the performance)

HAOCHENYE:
BTW, since `add_graph` has not been used by the downstream repo, could we rename it to a more proper name?

shenmishajing:
> Will it be more proper to call `add_graph` in `NativeVisualizerHook.before_train`, and this feature can be enabled by setting `watch_model=True` (default False, not sure will it harm the performance)

I agree with that. But, does the NativeVisualizerHook have been added to runner now? I do not find it now, and that's why I add it in runner init directly.

HAOCHENYE:
`NativeVisualizerHook` should be a custom hook. If users need to visualize something,  they can inherit or copy `NativeVisualizerHook` to implement custom `VisualizerHook`. The code snippet of config to enable `VisualizerHook` is like this:

`config_for_xxx.py`
```python
custom_hooks = [
    dict(type='NativeVisualizerHook')
    ...
]
```

shenmishajing:
> `NativeVisualizerHook` should be a custom hook. If users need to visualize something, they can inherit or copy `NativeVisualizerHook` to implement custom `VisualizerHook`. The code snippet of config to enable `VisualizerHook` is like this:
> 
> `config_for_xxx.py`
> 
> ```python
> custom_hooks = [
>     dict(type='NativeVisualizerHook')
>     ...
> ]
> ```

Ok, got it. I think it's more proper to call add_graph in `NativeVisualizerHook.before_train`.

HAOCHENYE:
I think `add_config` is a more commonly used function for the general experiments, it is ok to be called in `runner.__init__() `.

shenmishajing:
> 

Ok, thanks for your reply.

hhaAndroid (approved):

mmengine/visualization/vis_backend.py
```diff
@@ -360,19 +360,22 @@ class WandbVisBackend(BaseVisBackend):
             `wandb docs <https://docs.wandb.ai/ref/python/run#log_code>`_
             for details. Defaults to None.
             New in version 0.3.0.
+        watch_kwargs (optional, dict): Agurments for ``wandb.watch``.
```
zhouzaida, line 363:
```suggestion
        watch_kwargs (optional, dict): Agurments for ``wandb.watch``.
            New in version 0.3.0.
```

zhouzaida:
Hi @shenmishajing , thanks for your contribution. After having a discussion with WandB team, they have no plan to support visualizing models in `wanbd.watch` but we will merge this PR because it is a useful feature.

mmengine/visualization/vis_backend.py
```diff
@@ -360,19 +360,23 @@ class WandbVisBackend(BaseVisBackend):
...
+            New in version 0.3.0.
```
zhouzaida, line 364:
```suggestion
            New in version 0.4.0.
```

zhouzaida (approved):

HAOCHENYE (approved):

PR #665, opened 2022-11-02
[Fix] Fix potential OOM error when loading checkpoint with `map_location="cuda:0"`

HAOCHENYE:

## Motivation

If a checkpoint is saved with `map_location='cuda:0'`, `PretrainedInit` will first load the checkpoint to "cuda:0", and then copy the parameters to the model. During ddp-training, all ranks will first load the checkpoint to "cuda:0", which could cause an OOM error.

A workaround could be that let `PretrainedInit` load the checkpoint to "cpu" by default.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/model/weight_init.py
+++ b/mmengine/model/weight_init.py
@@ -472,7 +472,7 @@ class PretrainedInit:
         map_location (str): map tensors into proper locations.
     """
 
-    def __init__(self, checkpoint, prefix=None, map_location=None):
+    def __init__(self, checkpoint, prefix=None, map_location='cpu'):
         self.checkpoint = checkpoint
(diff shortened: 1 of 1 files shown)


mmengine/model/weight_init.py
```diff
@@ -472,7 +472,7 @@ class PretrainedInit:
         map_location (str): map tensors into proper locations.
```
C1rN09, line 472:
```suggestion
        map_location (str): map tensors into proper locations.
            Defaults to 'cpu'
```

zhouzaida (approved):

PR #690, opened 2022-11-07
[Enchance] remove redundant attribute in dump or print

Harold-lkk:
## Motivation

Remove the redundant inner attribute created by the property function

Before:
```python
class DetDataSample(BaseDataElement):

    @property
    def proposals(self):
        return self._proposals

    @proposals.setter
    def proposals(self, value):
        self.set_field(value=value, name='_proposals', dtype=BaseDataElement)

    @proposals.deleter
    def proposals(self):
        del self._proposals

data_sample = DetDataSample(proposals=InstanceData())

print(data_sample.keys())
```
```
['proposals', '_proposals']
```
```python
print(data_sample)
```
```
<DetDataSample(

    META INFORMATION

    DATA FIELDS
    proposals: <InstanceData(
        
            META INFORMATION
        
            DATA FIELDS
        ) at 0x7fa086437820>
    _proposals: <InstanceData(
        
            META INFORMATION
        
            DATA FIELDS
        ) at 0x7fa086437820>
) at 0x7fa086437910>
```

After:
```python
print(data_sample.keys())
```
```
['proposals']
```
```python
print(data_sample)
```
```
<DetDataSample(

    META INFORMATION

    DATA FIELDS
    proposals: <InstanceData(
        
            META INFORMATION
        
            DATA FIELDS
        ) at 0x7fa086437820>
) at 0x7fa086437910> 
```

### DIFF
--- a/mmengine/structures/base_data_element.py
+++ b/mmengine/structures/base_data_element.py
@@ -1,5 +1,7 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import copy
+import inspect
+import sys
 from typing import Any, Iterator, Optional, Tuple, Type, Union
 
 import numpy as np
@@ -458,7 +460,14 @@ class BaseDataElement:
                 raise AttributeError(
                     f'Cannot set {name} to be a field of data '
                     f'because {name} is already a metainfo field')
-            self._data_fields.add(name)
+            # The name only added to `data_fields`` when it is not the
+            # attribute related to property(methods decorated by @property).
+            if not isinstance(
+                    getattr(
+                        type(self),
+                        inspect.getframeinfo(sys._getframe(1)).function, None),
+                    property):
+                self._data_fields.add(name)
         super().__setattr__(name, value)
 
     # Tensor-like methods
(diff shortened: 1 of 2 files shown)


zhouzaida:
Please update unit tests.

HAOCHENYE (commented):
I think the key reason to save the duplicated attributes is that the calling stack of data_sample.gt_instance is:

`__setattr__` ->`set_fielld` -> `super().__setattr__` -> `set_field`

It is obvious that `set_filed` will be called twice, and I think we should avoid this behavior. After that, the redundant attributes will not be saved anymore.

mmengine/structures/base_data_element.py
```diff
@@ -458,7 +460,14 @@ def set_field(self,
...
                     f'Cannot set {name} to be a field of data '
                     f'because {name} is already a metainfo field')
-            self._data_fields.add(name)
+            # The name only added to `data_fields`` when it is not the
+            # attribute related to property(methods decorated by @property).
+            if not isinstance(
+                    getattr(
+                        type(self),
+                        inspect.getframeinfo(sys._getframe(1)).function, None),
+                    property):
```
mzr1996, line 469: Why not directly use
```suggestion
            # The private field won't be recorded by `data_fields`.
            if not name.starts_with('_'):
```

zhouzaida (approved):

PR #693, opened 2022-11-07
[Enhance] Support config torch enviroment in config.

HAOCHENYE:

## Motivation

Support config torch enviroment in config by the code block below:

```python
# Define the config
cfg = ...
cfg.env_cfg.torch_cfg = torch_cfg = dict(
    backends.cuda.matmul.allow_tf32=True,
    torch.backends.cuda.matmul.allow_fp16_reduced_precision_reduction=True
)
```
After building the runner, `torch.backends.cuda.matmul.allow_tf32` and `torch.backends.cuda.matmul.allow_fp16_reduced_precision_reduction` are set to True

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


mmengine/runner/runner.py
```diff
@@ -690,6 +690,41 @@ def set_randomness(self,
...
+        """Set torch variable to by config.
+
+        Args:
+            torch_cfg (dict): Contains key-value pair which defines the names
+                of torch variables and corresponding values.
+
+        Examples:
+            >>> torch_cfg = dict(
+            >>>    backends.cuda.matmul.allow_tf32=True,
+            >>>    torch.backends.cuda.matmul.allow_fp16_reduced_precision_reduction=True
```
RangiLyu, line 703: `torch.backends....` or `backends....`?

mmengine/runner/runner.py
```diff
@@ -690,6 +690,41 @@ def set_randomness(self,
...
+                    'should be set in `randomness_cfg`, rather than '
+                    '`torch_cfg`')
+            attributes_list = attributes.split('.')
+            modules = attributes_list[:-1]
+            attribute = attributes_list[-1]
+            target_module = torch
+            try:
+                for module in modules:
+                    target_module = getattr(target_module, module)
+                setattr(target_module, attribute, value)
```
RangiLyu, line 722: Is `setattr` suitable for all conditions?
How about using `eval(str)`?

mmengine/runner/runner.py
```diff
@@ -690,6 +690,41 @@ def set_randomness(self,
             deterministic=deterministic,
             diff_rank_seed=diff_rank_seed)
 
+    def set_torch_cfg(self, torch_cfg: dict) -> None:
```
RangiLyu, line 693: Can it be more general? Not only for torch?

mmengine/runner/runner.py
HAOCHENYE: both are OK, since you can access `torch.torch.torch...` recursively :rofl:

mmengine/runner/runner.py
HAOCHENYE: What's the scope of this function? If we want to support other environments not only for torch, and enable/disable some functions by calling some interfaces, we've to use `eval(str)`. But it also means we need to write the config like a code, is this what we expected?

C1rN09 (commented):
I have to say that `eval`/`exec` might make it unsafe. Besides, users might abuse it... Maybe we should consider twice on this?

mmengine/runner/runner.py
```diff
@@ -687,6 +687,43 @@ def set_randomness(self,
...
+                    'backends.cudnn.benchmark or backends.cudnn.deterministic '
+                    'should be set in `randomness_cfg`, rather than '
+                    '`torch_cfg`')
+            attributes_list = attributes.split('.')
+            modules = attributes_list[:-1]
+            attribute = attributes_list[-1]
+            target_module = torch
+            try:
+                for module in modules:
+                    target_module = getattr(target_module, module)
```
RangiLyu, line 720: the example docstring shows both `torch.xxx` and `xxx.xx` are supported. But this may cause `torch.torch` error.

RangiLyu (approved):

mmengine/runner/runner.py
```diff
@@ -687,6 +687,43 @@ def set_randomness(self,
...
+            >>>    torch.backends.cuda.matmul.allow_fp16_reduced_precision_reduction=True
+            >>>    )
+        """  # noqa: E501
+        for attributes, value in torch_cfg.items():
+            assert isinstance(attributes, str)
+            if attributes in [
+                    'backends.cudnn.benchmark', 'backends.cudnn.deterministic',
+                    'torch.backends.cudnn.benchmark',
+                    'torch.backends.cudnn.deterministic'
+            ]:
```
C1rN09, line 709: I think it's better to remove all `torch.` prefix in `attributes` and then check if it's in the first 2 options. Otherwise some abnormal configurations like `torch.torch.backends.cudnn.benchmark` may spoil the thing.

C1rN09 (approved):

PR #698, opened 2022-11-08
Fix `experiment_name` settings in runner.py

PingHGao:
Fix `experiment_name` settings in runner.py.

## Motivation

The `experiment_name` param in `Runner()` construction makes no scence, bucause a small mistake.
I correct this mistake here.

## Modification

I just change line 359 in runner.py from:
```python
self._log_dir = osp.join(self.work_dir, self.timestamp)
```
to
```python
self._log_dir = osp.join(self.work_dir, self.experiment_name)
```

## BC-breaking (Optional)

Nope

## Use cases (Optional)

Nope

## Checklist

### DIFF
--- a/mmengine/runner/runner.py
+++ b/mmengine/runner/runner.py
@@ -356,7 +356,7 @@ class Runner:
             self._experiment_name = f'{filename_no_ext}_{self._timestamp}'
         else:
             self._experiment_name = self.timestamp
-        self._log_dir = osp.join(self.work_dir, self.timestamp)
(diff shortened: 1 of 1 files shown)


zhouzaida:
Hi @PingHGao , thanks for your contribution. However, it is not a bug. The work_dir is the same as the experiment name, and the `log_dir` appends the timestamp to distinguish the resuming training case.

PingHGao:
ok, I got it.

Issue #688, opened 2022-11-07
Log issues when using two GPUs


PingHGao:
I have a training dataset which have 3425 items. When training with `batch_size = 4`, the log is:

```
2022/11/07 10:41:22 - mmengine - INFO - Epoch(train) [1][10/857]  lr: 1.0000e-04 
```
which means the length of the data_loader  857. This is correct. 
**But when I use 2 GPUs, the log remains the same!** It also says the data_loader's length is 857, which should be half because the batch_size is doubled with two GPUs. I can see that each GPU is processing 4 batch data through the GPU memory comsumption. 
Is there anything wrong with data_loader or logger? Please help!

PingHGao:
I have just tried the official example `examples/train.py`, which is the same.
Does the `examples/train.py` is just an example to show how to use two gpus and it do not provide the correct way to create a dataloader?

HAOCHENYE:
Thanks for your feedback! It seems that `examples/train.py` does not configure a distributed sampler for Dataloadera correctly, which contributes to the error indices in Dataloader. Would you mind trying  to configure the Dataloader with the [DefaultSampler](https://github.com/open-mmlab/mmengine/blob/ed20a9cba52ceb371f7c825131636b9e2747172e/mmengine/dataset/sampler.py#L14)  in `examples/train.py`? If it works, it could be helpful to post a PR to fix the example :smile:

PingHGao:
> Thanks for your feedback! It seems that `examples/train.py` does not configure a distributed sampler for Dataloadera correctly, which contributes to the error indices in Dataloader. Would you mind trying to configure the Dataloader with the [DefaultSampler](https://github.com/open-mmlab/mmengine/blob/ed20a9cba52ceb371f7c825131636b9e2747172e/mmengine/dataset/sampler.py#L14) in `examples/train.py`? If it works, it could be helpful to post a PR to fix the example smile

After configure the dataloader with `DefaultSampler`, the problem remains. The dataloader is as below. I find that the 2 GPU is running with the command `python -m torch.distributed.launch --nproc_per_node=2 examples/train.py --launcher pytorch`, but the code `print(get_dist_info())` gives `0, 1` in each process.
```
train_set = torchvision.datasets.CIFAR10(
            'data/cifar10',
            train=True,
            download=True,
            transform=transforms.Compose([
                transforms.RandomCrop(32, padding=4),
                transforms.RandomHorizontalFlip(),
                transforms.ToTensor(),
[...]

PR #696, opened 2022-11-07
[Doc] Add dataset & dataloader tutorial

C1rN09:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

Please describe the motivation of this PR and the goal you want to achieve through this PR.

## Modification

Please briefly describe what modification is made in this PR.

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repos?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #700, opened 2022-11-08
Update distributed_training.py

PingHGao:
Better example for DDP training

## Motivation

The previous version of `example/distributed_training.py` can not run DDP training correctly. After the discussion in #688，we provide a better example.

## Modification

Only `example/distributed_training.py` is modified, mainly  the dataloader. And I have done sufficient test.

## BC-breaking (Optional)

Nope

## Use cases (Optional)

Nope

## Checklist

### DIFF
--- a/examples/distributed_training.py
+++ b/examples/distributed_training.py
@@ -1,11 +1,11 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import argparse
 
+import torch
 import torch.nn.functional as F
 import torchvision
 import torchvision.transforms as transforms
 from torch.optim import SGD
-from torch.utils.data import DataLoader
 
 from mmengine.evaluator import BaseMetric
 from mmengine.model import BaseModel
@@ -39,6 +39,14 @@ class Accuracy(BaseMetric):
         total_correct = sum(item['correct'] for item in results)
         total_size = sum(item['batch_size'] for item in results)
         return dict(accuracy=100 * total_correct / total_size)
+    
+    
+class IntToTensor:
+    def __call__(self, data):
+        return torch.tensor(data)
+
+    def __repr__(self):
+        return self.__class__.__name__ + '()'
 
 
 def parse_args():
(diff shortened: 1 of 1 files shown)


examples/distributed_training.py
```diff
@@ -41,6 +41,14 @@ def compute_metrics(self, results):
         return dict(accuracy=100 * total_correct / total_size)
 
 
+class IntToTensor:
+    def __call__(self, data):
+        return torch.tensor(data)
+
+    def __repr__(self):
+        return self.__class__.__name__ + '()'
+
```
C1rN09, line 50: This target_transform is not necessary here. Maybe you encountered error caused by MMEngine's default `collate_fn`? Our new [documentation](https://github.com/open-mmlab/mmengine/pull/696/files) (which is still WIP) detailed on this and you may give some feedback :smile:

examples/distributed_training.py
```diff
@@ -57,29 +65,37 @@ def parse_args():
...
-        dataset=torchvision.datasets.CIFAR10(
-            'data/cifar10',
-            train=False,
-            download=True,
-            transform=transforms.Compose(
-                [transforms.ToTensor(),
-                 transforms.Normalize(**norm_cfg)])))
+        dataset=valid_set,
+        sampler=dict(type='DefaultSampler', shuffle=True),
+        collate_fn=dict(type='default_collate'))
```
C1rN09, line 98: I think in this example, it's better to remove `shuffle` parameter in  the outer `dict` because it might be misleading. And it might be better to change `shuffle` in `val_dataloader` to `False`, which is a common practice.

PingHGao:
Thanks for your review, I have updated the file and tested. @C1rN09

C1rN09 (dismissed):

C1rN09:
Sorry I approved by mistake. Your lint failed and you can fix it by using `pre-commit hooks` following this [guide](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md#code-style)

Or to be simple, you can run these commands in mmengine project root directory

```shell
pip install -U pre-commit
pre-commit install
pre-commit run --all-files
```

C1rN09:
I see you change it back to the original wrong version... Is it by mistake?

PingHGao:
> I see you change it back to the original wrong version... Is it by mistake?

Sorry, I missed up the main branch. I think this update should be allright.

C1rN09 (approved):

PR #723, opened 2022-11-15
[Enhance] Support doing validation on non-ema model.

mzr1996:
## Motivation

Sometimes, we want to do validation on the non-ema model although the EMA is enabled.

## Modification

Add an option to determine whether to do validation on EMA models.

## BC-breaking (Optional)

No

## Use cases (Optional)

```python
custom_hooks = [dict(type='EMAHook', validate_on_ema=False)]
```

## Checklist


zhouzaida:
closed after discussion with @mzr1996

PR #739, opened 2022-11-17
[Feature] Support EarlyStoppingHook

### LINKED ISSUE #356: Support Early Stop
**Describe the feature**

**Motivation**
See an issue in https://github.com/open-mmlab/mmdetection/issues/1383. May stop early stop in mmengine so that all OpenMMLab repos could use this feature.

**Related resources**
If there is an official code release or third-party implementations, please also provide the information here, which would be very helpful.

**Additional context**
If you would like to implement the feature and create a PR, please leave a comment here and that would be much appreciated.

nijkah:
## Motivation
Closes #356 

## Modification
Added a new variable `stop_training` in `EpochBasedTrainLoop` and `IterBasedTrainLoop`.
Edited `run` method as
```python
    def run(self) -> torch.nn.Module:
        """Launch training."""
        while self._epoch < self._max_epochs and not self.stop_training:
            self.run_epoch()
```

**
`EarlyStoppingHook` itself does not save the best checkpoint. Should we support it?

I referred to some logics from
- https://github.com/Bjarten/early-stopping-pytorch/blob/master/pytorchtools.py
- https://docs.ray.io/en/latest/_modules/ray/tune/stopper/experiment_plateau.html#ExperimentPlateauStopper


mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,98 @@
...
+        compare = self.rule_map[self.rule]
+
+        self.pool_values.append(latest_value)
+
+        if self.rule == 'greater':
+            # maintain largest values
+            self.pool_values = sorted(self.pool_values)[-self.pool_size:]
+        else:
+            # maintain smalleast values
+            self.pool_values = sorted(self.pool_values)[:self.pool_size]
```
HAOCHENYE, line 84: How about using `PriorityQueue`?

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,98 @@
...
+    Args:
+        metric (str): The metric key to decide early stopping.
+        rule (str, optional): Comparison rule. Options are 'greater',
+            'less'. Defaults to None.
+        delta(float, optional): Minimum difference to continue the training.
+            Defaults to 0.01.
+        pool_size (int, optional): The number of experiments to consider.
+            Defaults to 5.
+        patience (int, optional): Maximum number of tolerance.
+            Defaults to 0.
```
HAOCHENYE, line 24: Hi, referring to https://github.com/Bjarten/early-stopping-pytorch/blob/master/pytorchtools.py
Training will be stopped when the metric gets consistently worse "patience" times. The `rule` argument means the better trends we want, and the patience means How long to wait after last time validation metric improved. Such setting makes sense to me.

If I understand correctly, the current implementation will stop training when the metric gets consistently better "patience" times, is it our expected?

HAOCHENYE: BTW, I think we could also adopt the strategy here https://github.com/open-mmlab/mmengine/pull/739#pullrequestreview-1185414869

If we reach the target metric threshold, just stop training.

nijkah: Inspired by [`ray.tune.stopper`](https://docs.ray.io/en/latest/_modules/ray/tune/stopper/experiment_plateau.html#ExperimentPlateauStopper),
my original intention was "Stop training when the current score does not exceed "the delta plus average of top `pool_size` scores. But it is acceptable to `patience` times when the conditions are not met.".

In this aspect, `rule` means that "current score should be `'greater|less'` than the average of latest scores.".
So, to check when the condition does not meet, we put the 'the average of latest scores' as a left argument.
```python
compare(sum(self.pool_values) / self.pool_size + self.delta, latest_value)
```

To my understanding, the current implementation meets your description. Is it better to improve the clarity?

mmengine/hooks/early_stopping_hook.py
nijkah: https://stackoverflow.com/a/26441285
If `rule=='greater`, getting values from the `PriorityQueue` should work differently. `Q.get()` or `Q.get()[1]`.
Is there any good idea to handle this case?

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,98 @@
...
+            '`rule` should be either \'greater\' or \'less\'.'
+        self.rule = rule
+        self.delta = delta
+        self.pool_size = pool_size
+        self.patience = patience
+        self.count = 0
+
+        self.pool_values: List[float] = []
+
+    def after_val_epoch(self, runner, metrics):
```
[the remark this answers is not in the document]
nijkah: Added it!

mmengine/hooks/early_stopping_hook.py
HAOCHENYE: Initially, I thought we can put + or - metrics in `PriorityQueue` to represent `greater` or `less`. Emmmm. But I think since the metrics list will not be too long, it could be OK to directly use `list` rather than `PriorityQueue`

mmengine/hooks/early_stopping_hook.py
HAOCHENYE: Ah, I don't know what happened, I misunderstood in the morning :laughing: ,my bad :rofl: 

But maybe we should set the default value of patience to 1 or a larger one? I think 0 or 1 is too strict, and the model will be very easy to trigger the early stop.

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,108 @@
...
+
+DATA_BATCH = Optional[Union[dict, tuple, list]]
+
+
+@HOOKS.register_module()
+class EarlyStoppingHook(Hook):
+    """Early stop the training when the metric reached a plateau.
+
+    Args:
+        metric (str): The metric key to decide early stopping.
```
HAOCHENYE, line 16: Referring to https://keras.io/api/callbacks/early_stopping/, maybe `monitor` will be a more proper parameter name? Maybe someday MMEngine will support calculate the validation loss in the train loop :rofl:

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,108 @@
...
+        rule (str, optional): Comparison rule. Options are 'greater',
+            'less'. Defaults to None.
+        delta(float, optional): Minimum difference to continue the training.
+            Defaults to 0.01.
+        pool_size (int, optional): The number of experiments to consider.
+            Defaults to 5.
+        patience (int, optional): Maximum number of tolerance.
```
HAOCHENYE, line 23: Ahhh, it could be better to describe the patience in more detailed

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,108 @@
...
+
+        if self.rule == 'greater':
+            # maintain largest values
+            self.pool_values = sorted(self.pool_values)[-self.pool_size:]
+        else:
+            # maintain smalleast values
+            self.pool_values = sorted(self.pool_values)[:self.pool_size]
+
+        if len(self.pool_values) == self.pool_size and compare(
+                sum(self.pool_values) / self.pool_size + self.delta,
```
HAOCHENYE, line 97: Should delta be "added" or "minors" according to the rule?

mmengine/hooks/early_stopping_hook.py
nijkah: I borrowed the description from https://docs.ray.io/en/latest/_modules/ray/tune/stopper/experiment_plateau.html#ExperimentPlateauStopper.

Is the description from https://keras.io/api/callbacks/early_stopping/ better?

nijkah:
Hi @HAOCHENYE, I have a little question.
In the UT with initializing `Runner`, the error raises when the `experiment_name` of `Runner` isn't specified. Also, it happens when the `experiment_name` is duplicated with the other UTs although there is a logic for cleaning its temporary directory.

Is it intended (the developers should consider) or unintended behavior?

HAOCHENYE:
> Hi @HAOCHENYE, I have a little question. In the UT with initializing `Runner`, the error raises when the `experiment_name` of `Runner` isn't specified. Also, it happens when the `experiment_name` is duplicated with the other UTs although there is a logic for cleaning its temporary directory.
> 
> Is it intended (the developers should consider) or unintended behavior?

Usually, if the user does not specify the experiment name, `Runner` will automatically generate the experiment name and each `Runner` instance will have unique experiment names **with different timestamp suffix**. However, if the time interval between `Runner`'s creation is so short that the timestamp suffixes are identical, it could lead to Runners with duplicate experiment names, which is not allowed in current design.  

MMEngine has multiple global variances such as `MessageHub`,  `Visualizer` and `DefaultScope`, and all these variables are marked with the experiment name of `Runner`. Global variables are very dangerous, therefore these variables must have unique names.

As for the unit tests will often raise an error about the duplicated experiment name, it is caused by the irregularity of the current unit testing, and we plan to refactor the unit tests, and provide a more general function to build Runner in unit tests.

mmengine/hooks/early_stopping_hook.py
HAOCHENYE: LGTM 😆!

mmengine/hooks/early_stopping_hook.py
HAOCHENYE: Yep, I think the description of https://keras.io/api/callbacks/early_stopping/ is better. BTW, the `epoch` in the description is not accurate, will `the times of validation` be better?

mmengine/hooks/early_stopping_hook.py
nijkah: I'm not sure about the naming. But, I think `delta` is still intuitive to understand what it means.
Then, how about `min_delta`, which means `minimum difference that must be appeared` as in https://keras.io/api/callbacks/early_stopping/?

nijkah:
One more question! I am thinking about which experiments we should consider for no improvement.

Current implementation retains **top scores** ignoring the recent inferior results.
But it may be more intuitive to retain **latest scores** to check the trend of training.

Which one will be better?

HAOCHENYE:
> One more question! I am thinking about which experiments we should consider for no improvement.
> 
> Current implementation retains **top scores** ignoring the recent inferior results. But it may be more intuitive to retain **latest scores** to check the trend of training.
> 
> Which one will be better?

Additional notes： We could also borrow the implementation in 

https://github.com/Lightning-AI/lightning/blob/9a4e8a8c528d6475ab33c46a0a84e27273cc10bd/src/pytorch_lightning/callbacks/early_stopping.py#L38

Back to the topic, it could make more sense to me to directly compare the latest metric with the previous best Nth metrics and finally update the Nth metrics with the latest one, rather than first update the Nth metrics, and then compare them 🤣. what do you think about it?

nijkah:
I think your provided link looks neat. I'll rewrite the code especially for stopping logic and `__init__` parameters.

mmengine/hooks/__init__.py
```diff
@@ -12,7 +13,16 @@
...
+    'DistSamplerSeedHook',
+    'ParamSchedulerHook',
+    'SyncBuffersHook',
+    'EmptyCacheHook',
+    'CheckpointHook',
+    'LoggerHook',
+    'NaiveVisualizationHook',
+    'EMAHook',
+    'RuntimeInfoHook',
+    'EarlyStoppingHook',
```
zhouzaida, line 27:
```suggestion
    'EarlyStoppingHook'
```

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,140 @@
...
+        stopping_threshold: Optional[float] = None,
+    ):
+
+        self.monitor = monitor
+        if monitor in self._default_greater_keys:
+            rule = 'greater'
+        elif monitor in self._default_less_keys:
+            rule = 'less'
+        assert rule in ['greater', 'less'], \
+            '`rule` should be either \'greater\' or \'less\'.'
```
zhouzaida, line 57:
```suggestion
            "`rule` should be either 'greater' or 'less'."
```

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,140 @@
...
+        self.default_end_message = ' This training process stops early.'
+
+    def _check_stop_condition(self, current_score):
+        compare = self.rule_map[self.rule]
+        stop_training = False
+
+        if self.check_finite and not isfinite(current_score):
+            stop_training = True
+            self.reason_message = (f'Monitored metric {self.monitor} = '
+                                   f'{current_score} is not finite. '
```
zhouzaida, line 77:
```suggestion
                                   f'{current_score} is infinite. '
```

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,140 @@
...
+from typing import Optional, Union
+
+from mmengine.registry import HOOKS
+from .hook import Hook
+
+DATA_BATCH = Optional[Union[dict, tuple, list]]
+
+
+@HOOKS.register_module()
+class EarlyStoppingHook(Hook):
```
zhouzaida, line 13: Consider recording the state of early stopping to support resuming training.

mmengine/runner/loops.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         self._iter = 0
         self.val_begin = val_begin
         self.val_interval = val_interval
+        self.stop_training = False
```
zhouzaida, line 52:
```suggestion
        # This attribute will be updated by `EarlyStoppingHook`
        # when it is enabled.
        self.stop_training = False
```

mmengine/runner/loops.py
```diff
@@ -216,6 +217,7 @@ def __init__(
...
```
zhouzaida, line 220:
```suggestion
        # This attribute will be updated by `EarlyStoppingHook`
        # when it is enabled.
        self.stop_training = False
```

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,140 @@
...
+                raise RuntimeError(
+                    f'Early stopping conditioned on metric '
+                    f'`{self.monitor} is not available. Please check available'
+                    f' metrics {metrics}, or set `strict=False` in '
+                    f'`EarlyStoppingHook`.')
+            warnings.warn(
+                f'Skip early stopping process since the evaluation '
+                f'results ({metrics.keys()}) do not include `monitor` '
+                f'({self.monitor}).')
+            return
```
zhouzaida, line 134:
```suggestion
        if self.monitor not in metrics:
            if self.strict:
                raise RuntimeError(
                    'Early stopping conditioned on metric '
                    f'`{self.monitor} is not available. Please check available'
                    f' metrics {metrics}, or set `strict=False` in '
                    '`EarlyStoppingHook`.')
            warnings.warn(
                'Skip early stopping process since the evaluation '
                f'results ({metrics.keys()}) do not include `monitor` '
                f'({self.monitor}).')
            return
```

zhouzaida:
Please also append the Hook in the following files.

- https://github.com/open-mmlab/mmengine/blob/main/docs/zh_cn/api/hooks.rst
- https://github.com/open-mmlab/mmengine/blob/main/docs/en/api/hooks.rst

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,140 @@
...
+        stopping_threshold: Optional[float] = None,
+    ):
+
+        self.monitor = monitor
+        if monitor in self._default_greater_keys:
+            rule = 'greater'
+        elif monitor in self._default_less_keys:
+            rule = 'less'
+        assert rule in ['greater', 'less'], \
+            "`rule` should be either 'greater' or 'less'."
```
HAOCHENYE, line 57:
```suggestion
            '`rule` should be either "greater" or "less".'
```

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,140 @@
...
+        self.rule = rule
+        self.min_delta = min_delta
+        self.strict = strict
+        self.check_finite = check_finite
+        self.patience = patience
+        self.stopping_threshold = stopping_threshold
+
+        self.wait_count = 0
+        self.best_score = -inf if rule == 'greater' else inf
+        self.reason_message = ''
```
HAOCHENYE, line 67: Suggest directly return `reason_message` and `stop_training` in `_check_stop_condition`.

mmengine/hooks/early_stopping_hook.py
HAOCHENYE: `MessageHub` will save all history metrics during training, maybe we could utilize it to resume training

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,139 @@
...
+        self.rule = rule
+        self.min_delta = min_delta
+        self.strict = strict
+        self.check_finite = check_finite
+        self.patience = patience
+        self.stopping_threshold = stopping_threshold
+
+        self.wait_count = 0
+        self.best_score = -inf if rule == 'greater' else inf
+        self.default_end_message = ' This training process stops early.'
```
HAOCHENYE, line 67: `self.default_end_message` does not need to be an attribute anymore.

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def get_mock_runner():
+    runner = Mock()
+    runner.train_loop = Mock()
+    runner.train_loop.stop_training = False
+    return runner
+
+
+class TestEarlyStoppingHook(TestCase):
```
HAOCHENYE, line 85: Hi! You can use `RunnerTestCase` defined [here](https://github.com/open-mmlab/mmengine/blob/main/mmengine/testing/runner_test_case.py) to test hooks :smile:

HAOCHENYE (commented):
Verrrrrrry thank you for the contributions :laughing: ! This pr will be ready to merge after supporting resuming and updating the unit test :rocket:  :rocket:! Very much looking forward to this new feature!

mmengine/hooks/early_stopping_hook.py
nijkah: @HAOCHENYE I'll continue this work this week! Sorry for late response 😢

zhouzaida:
Hi @nijkah , is there any progress?

nijkah:
@zhouzaida Truly apologize for the late response. I think I can start working on it by this Wednesday.

zhouzaida:
Hi @nijkah , is there any progress?

nijkah:
@zhouzaida @HAOCHENYE  Sorry and thanks for reminding me. I'll definitely continue this until through the holiday week.

nijkah:
Hi @HAOCHENYE , really apologize, and thanks for your patience. 
Now, I am thinking about supporting resuming logic.
My initial idea is just to use `save_last` flag and borrow the logic of `CheckpointHook._save_checkpoint(runner)`.

You already mentioned the `MessageHub`, can I ask you to give some guide on how to utilize `MessageHub` to support resuming?

HAOCHENYE:
> Hi @HAOCHENYE , really apologize, and thanks for your patience. Now, I am thinking about supporting resuming logic. My initial idea is just to use `save_last` flag and borrow the logic of `CheckpointHook._save_checkpoint(runner)`.
> 
> You already mentioned the `MessageHub`, can I ask you to give some guide on how to utilize `MessageHub` to support resuming?

Hi! History information is saved in `MessageHub`, such as loss history and metric history. These variables are updated by `MessageHub.update_scalar` like this:

https://github.com/open-mmlab/mmengine/blob/6dc1d7082a1e2bc32049e65f283c4fd20de1c026/mmengine/hooks/runtime_info_hook.py#L120

Then, these variables can be resumed from `runner.message_hub`. Maybe EarlyStoppingHook can refer to this way?

zhouzaida:
Hi @nijkah , thanks for your contribution. We are planning to merge this PR today and release a new version 0.6.0. The feature to resume can be supported in the future if users require it.

mmengine/hooks/__init__.py
```diff
@@ -17,5 +18,5 @@
     'Hook', 'IterTimerHook', 'DistSamplerSeedHook', 'ParamSchedulerHook',
     'SyncBuffersHook', 'EmptyCacheHook', 'CheckpointHook', 'LoggerHook',
     'NaiveVisualizationHook', 'EMAHook', 'RuntimeInfoHook', 'ProfilerHook',
-    'NPUProfilerHook', 'PrepareTTAHook'
+    'PrepareTTAHook', 'EarlyStoppingHook', 'NPUProfilerHook',
```
zhouzaida, line 21:
```suggestion
    'PrepareTTAHook', 'NPUProfilerHook', 'EarlyStoppingHook'
```

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,146 @@
...
+
+        self.monitor = monitor
+        if monitor in self._default_greater_keys:
+            rule = 'greater'
+        elif monitor in self._default_less_keys:
+            rule = 'less'
+        assert rule in ['greater', 'less'], \
+            '`rule` should be either "greater" or "less".'
+        self.rule = rule
+        self.min_delta = min_delta
```
zhouzaida, line 62:
```suggestion
        self.min_delta = min_delta if rule == 'greater' else -1 * min_delta
```

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,146 @@
...
+        strict (bool, optional): Whether to crash the training when `monitor`
+            is not found in the `metrics`. Defaults to False.
+        check_finite: Whether to stop training when the monitor becomes NaN or
+            infinite. Defaults to True.
+        patience (int, optional): The times of validation with no improvement
+            after which training will be stopped. Defaults to 5.
+        stopping_threshold (float, optional): Stop training immediately once
+            the monitored quantity reaches this threshold. Defaults to None.
+        save_last (bool, optional): Whether to save last status when early
+            stopping is enabled.
```
zhouzaida, line 31:
```suggestion
            stopping is enabled.
            
       Note:
           `New in version 0.6.0.`
```

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,146 @@
...
+
+@HOOKS.register_module()
+class EarlyStoppingHook(Hook):
+    """Early stop the training when the monitored metric reached a plateau.
+
+    Args:
+        monitor (str): The monitored metric key to decide early stopping.
+        rule (str, optional): Comparison rule. Options are 'greater',
+            'less'. Defaults to None.
+        min_delta(float, optional): Minimum difference to continue the
```
zhouzaida, line 20:
```suggestion
        min_delta (float, optional): Minimum difference to continue the
```

nijkah:
> Hi @nijkah , thanks for your contribution. We are planning to merge this PR today and release a new version 0.6.0. The feature to resume can be supported in the future if users require it. In addition, this PR can be merged after resolving the conflict.

@zhouzaida I apologize for not finishing the rest work. I think it is okay to merge right now. I'll try to stay tuned to the future work for this.

zhouzaida:
> > Hi @nijkah , thanks for your contribution. We are planning to merge this PR today and release a new version 0.6.0. The feature to resume can be supported in the future if users require it. In addition, this PR can be merged after resolving the conflict.
> 
> @zhouzaida I apologize for not finishing the rest work. I think it is okay to merge right now. I'll try to stay tuned to the future work for this.

Hi @nijkah , thanks for your contribution again 👍.

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,149 @@
...
+
+        current_score = metrics[self.monitor]
+
+        stop_training, message = self._check_stop_condition(current_score)
+        if stop_training:
+            runner.train_loop.stop_training = True
+            runner.logger.info(message)
+            if self.save_last:
+                runner.logger.info(
+                    f'Saving last checkpoint at {runner.epoch + 1} epochs.')
```
zhouzaida, line 148:
```suggestion
                    f'Saving last checkpoint at {runner.epoch} epochs.')
```

zhouzaida: `runner.epoch` will increase by one after training an epoch that is ahead of validation. https://github.com/open-mmlab/mmengine/blob/d8abf9ac0ff053c2fc23f3902edb9aabebbe1ebf/mmengine/runner/loops.py#L109

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,245 @@
...
+        runner = get_mock_runner()
+        metrics = [{'accuracy/top1': i / 9.} for i in range(8)]
+        hook = EarlyStoppingHook(monitor='accuracy/top1', rule='greater')
+        for metric in metrics:
+            hook.after_val_epoch(runner, metric)
+            if runner.train_loop.stop_training:
+                break
+        self.assertAlmostEqual(hook.best_score, 7 / 9)
+
+        # Check smalleast value
```
zhouzaida, line 156:
```suggestion
        # Check smallest value
```

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,149 @@
...
+        strict (bool, optional): Whether to crash the training when `monitor`
+            is not found in the `metrics`. Defaults to False.
+        check_finite: Whether to stop training when the monitor becomes NaN or
+            infinite. Defaults to True.
+        patience (int, optional): The times of validation with no improvement
+            after which training will be stopped. Defaults to 5.
+        stopping_threshold (float, optional): Stop training immediately once
+            the monitored quantity reaches this threshold. Defaults to None.
+        save_last (bool, optional): Whether to save last status when early
+            stopping is enabled.
```
zhouzaida, line 31: This feature can be removed from the current version because it only prints a message.

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,141 @@
...
+        check_finite: bool = True,
+        patience: int = 5,
+        stopping_threshold: Optional[float] = None,
+    ):
+
+        self.monitor = monitor
+        if monitor in self._default_greater_keys:
+            rule = 'greater'
+        elif monitor in self._default_less_keys:
+            rule = 'less'
```
zhouzaida, line 58:
```suggestion
        greater_keys = [key.lower() for key in self._default_greater_keys]
        less_keys = [key.lower() for key in self._default_less_keys]
        if monitor.lower() in greater_keys:
            rule = 'greater'
        elif monitor.lower() in less_keys:
            rule = 'less'
```

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def test_init(self):
+
+        hook = EarlyStoppingHook(monitor='acc')
+        assert hook.rule == 'greater'
+        assert hook.best_score < 0
+
+        hook = EarlyStoppingHook(monitor='loss')
+        assert hook.rule == 'less'
+        assert hook.best_score > 0
```
zhouzaida, line 105:
```suggestion
        hook = EarlyStoppingHook(monitor='acc')
        assert hook.rule == 'greater'
        assert hook.best_score < 0

        hook = EarlyStoppingHook(monitor='ACC')
        assert hook.rule == 'greater'
        assert hook.best_score < 0        
        

        hook = EarlyStoppingHook(monitor='loss')
        assert hook.rule == 'less'
        assert hook.best_score > 0
        
        hook = EarlyStoppingHook(monitor='Loss')
        assert hook.rule == 'less'
        assert hook.best_score > 0
```

zhouzaida (approved):

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,268 @@
...
+        # `FileHandler` should be closed in Windows, otherwise we cannot
+        # delete the temporary directory
+        logging.shutdown()
+        MMLogger._instance_dict.clear()
+        self.temp_dir.cleanup()
+
+    def test_init(self):
+
+        hook = EarlyStoppingHook(monitor='acc')
+        assert hook.rule == 'greater'
```
HAOCHENYE, line 100: Suggest using `self.assertEqual` defined in `unittest.TestCase` to replace with `assert`

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,268 @@
...
+        # `train_loop` must contain `stop_training` variable.
+        with pytest.raises(AssertionError):
+            hook = EarlyStoppingHook(monitor='accuracy/top1', rule='greater')
+            hook.before_run(runner)
+
+    def test_after_val_epoch(self):
+        runner = get_mock_runner()
+
+        # if `monitor` does not match, skip the hook.
+        with pytest.warns(UserWarning) as record_warnings:
```
HAOCHENYE, line 146: Suggese using `self.assertWarns
`

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,268 @@
...
+
+        with pytest.raises(ValueError):
+            EarlyStoppingHook(monitor='accuracy/top1', rule='the world')
+
+    def test_before_run(self):
+        runner = Mock()
+        runner.train_loop = object()
+
+        # `train_loop` must contain `stop_training` variable.
+        with pytest.raises(AssertionError):
```
HAOCHENYE, line 138: Suggest using `self.assertRaises`

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,268 @@
...
+            f'Skip early stopping process since the evaluation results '
+            f'({metrics.keys()}) do not include `monitor` ({hook.monitor}).')
+        for warning in record_warnings:
+            if str(warning.message) == expected_message:
+                break
+        else:
+            assert False
+
+        # if `monitor` does not match and strict=True, crash the training.
+        with pytest.raises(RuntimeError):
```
HAOCHENYE, line 163: Suggest using self.assertRaises

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,268 @@
...
+        assert hook.rule == 'greater'
+        assert hook.best_score < 0
+
+        hook = EarlyStoppingHook(monitor='ACC')
+        assert hook.rule == 'greater'
+        assert hook.best_score < 0
+
+        hook = EarlyStoppingHook(monitor='ACC')
+        assert hook.rule == 'greater'
+        assert hook.best_score < 0
```
HAOCHENYE, line 109: Duplicated?

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,268 @@
...
+
+        hook = EarlyStoppingHook(monitor='Loss')
+        assert hook.rule == 'less'
+        assert hook.best_score > 0
+
+        hook = EarlyStoppingHook(monitor='ce_loss')
+        assert hook.rule == 'less'
+        assert hook.best_score > 0
+
+        with pytest.raises(ValueError):
```
HAOCHENYE, line 127: Suggest using `self.assertRaises`

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,268 @@
...
+            EarlyStoppingHook(monitor='recall')
+
+        with pytest.raises(ValueError):
```
HAOCHENYE, line 130: Suggest using `self.assertRaises`

HAOCHENYE: The rule `the world` reminds me of the comic [jojo](https://jojowiki.com/The_World)(one of my favorite comic :laughing: ). Just missing this comment If I misunderstand it :rofl:

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,268 @@
...
+        # Check stop training
+        runner = get_mock_runner()
+        metrics = [{'accuracy/top1': i} for i in torch.linspace(98, 99, 8)]
+        hook = EarlyStoppingHook(
+            monitor='accuracy/top1', rule='greater', min_delta=1)
+        for metric in metrics:
+            hook.after_val_epoch(runner, metric)
+            if runner.train_loop.stop_training:
+                break
+        assert runner.train_loop.stop_training
```
HAOCHENYE, line 198: `self.assertTrue`

tests/test_hooks/test_early_stopping_hook.py
```diff
@@ -0,0 +1,268 @@
...
+        # Check finite
+        runner = get_mock_runner()
+        metrics = [{'accuracy/top1': math.inf} for i in range(5)]
+        hook = EarlyStoppingHook(
+            monitor='accuracy/top1', rule='greater', min_delta=1)
+        for metric in metrics:
+            hook.after_val_epoch(runner, metric)
+            if runner.train_loop.stop_training:
+                break
+        assert runner.train_loop.stop_training
```
HAOCHENYE, line 209: Same as above

mmengine/hooks/early_stopping_hook.py
```diff
@@ -0,0 +1,159 @@
...
+            is not found in the `metrics`. Defaults to False.
+        check_finite: Whether to stop training when the monitor becomes NaN or
+            infinite. Defaults to True.
+        patience (int, optional): The times of validation with no improvement
+            after which training will be stopped. Defaults to 5.
+        stopping_threshold (float, optional): Stop training immediately once
+            the monitored quantity reaches this threshold. Defaults to None.
+
+       Note:
+           `New in version 0.6.0.`
```
zhouzaida, line 32:
```suggestion
           `New in version 0.7.0.`
```

zhouzaida (approved):

Issue #627, opened 2022-10-18
[Feature Request] Support deepspeed integration


nijkah:
**Describe the feature**

**Motivation**
Nowadays, [deepspeed](https://github.com/microsoft/DeepSpeed) became a fundamental framework that facilitates training and inference for large-scale or foundation models.
We are developing a feature for deepspeed integration into mmengine with support for a deepspeed-specified runner and optim_wrapper.

Does MMEngine have a plan to support deepspeed?
Then we can contribute to MMEngine with our implementation :)

Please let me know any guide, plan or opinion about this. :)

C1rN09:
Hi, @nijkah We welcome any kind of contribution, and deepspeed integration **is definitely** what we desire!
However, could you make it clearer about "deepspeed-specified runner and optim_wrapper"? If you are going to write a new `runner` that only serves deepspeed models, it seems not quite reasonable and we might need more discussion on it ^_^

C1rN09:
Hi, @nijkah Have you got any new progress on deepspeed integration? Hope we can discuss on it before you post a PR because it might not be a small & easy one. If you have any ideas/problems/progress, we are always open to have a discussion, either in this issue, or our [discussion board](https://github.com/open-mmlab/mmengine/discussions).

nijkah:
Hi, @C1rN09. Our integration development is almost done although there are still several choices left to consider.

Our current implementation supports
1. Enable ZeRO1, ZeRO2, ZeRO3
2. Saving a monolithic weight logic (DeepSpeed saves its model weights and optimizer's state in separate files, and the number of saved files are multiplied by the world_size.)

doesn't support yet
1. FP16 (There is a method to support it! But the solution is quite messy.)
2. Mixture of Experts
3. Pipeline Parallelism (It requires the logic to sequentialize `MM` models.
----

There are several reasons why we try to write a new deepspeed-dedicated runner.
Although we try to follow most of mmengine's `Runner` logic, there should be some modifications to support deepspeed.

Main logic of `DeepSpeedRunner` is like below,
```python
        >>> self.model = self.build_model(model)
        >>> self.optim_wrapper = self.build_optim_wrapper(optim_wrapper)
        >>> ds_config = json.load(open(cfg.deepspeed_config))
        >>> self.model, optimizer = deepspeed.initialize(
        >>>    model=self.model,
        >>>    optimizer=self.optim_wrapper.optimizer,
        >>>    model_parameters=self.model.parameters(),
[...]

PR #742, opened 2022-11-18
[Feature] Add ApexOptimWrapper

xcnick:

## Motivation

## Modification

Add ApexOptimWrapper for mmengine.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/optim/optimizer/apex_optimizer_wrapper.py
+++ b/mmengine/optim/optimizer/apex_optimizer_wrapper.py
@@ -74,8 +74,9 @@ class ApexOptimWrapper(OptimWrapper):
         If you use ``IterBasedRunner`` and enable gradient accumulation,
         the original `max_iters` should be multiplied by
         ``accumulative_counts``.
-        
-    `New in version 0.6.0.`
+
+    Note:
+        `New in version 0.6.0.`
     """  # noqa: E501
 
     def __init__(self,
--- a/mmengine/optim/optimizer/apex_optimizer_wrapper.py
+++ b/mmengine/optim/optimizer/apex_optimizer_wrapper.py
@@ -1,5 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 from contextlib import contextmanager
+from typing import Union
 
 import torch
 import torch.nn as nn
@@ -23,11 +24,17 @@ class ApexOptimWrapper(OptimWrapper):
     as ``OptimWrapper``.
 
     Warnings:
-        ``ApexOptimWrapper`` requires
-            [nvidia apex](https://github.com/NVIDIA/apex).
+        ``ApexOptimWrapper`` requires `nvidia apex
+        <https://github.com/NVIDIA/apex>`_
 
     Args:
 
+        opt_level (str, default="O1"): Pure or mixed precision
+            optimization level. Accepted values are "O0", "O1", "O2",
+            and "O3".
+        loss_scale (float or str, default=None): If passed as
+            a string, must be a string representing a number,
+            e.g., "128.0", or the string "dynamic".
         **kwargs: Keyword arguments passed to OptimWrapper.
 
     Note:
@@ -36,7 +43,10 @@ class ApexOptimWrapper(OptimWrapper):
         ``accumulative_counts``.
     """
 
-    def __init__(self, opt_level='O1', loss_scale='dynamic', **kwargs):
+    def __init__(self,
+                 opt_level: str = 'O1',
+                 loss_scale: Union[float, str] = 'dynamic',
+                 **kwargs):
         assert apex_amp is not None, \
             'Apex is not installed. Please check ' \
             'https://github.com/NVIDIA/apex#linux.'
@@ -44,7 +54,7 @@ class ApexOptimWrapper(OptimWrapper):
         self.opt_level = opt_level
         self.loss_scale = loss_scale
 
-    def backward(self, loss: torch.Tensor, **kwargs):
+    def backward(self, loss: torch.Tensor, **kwargs) -> None:
         """Perform gradient back propagation with :attr:`loss_scaler`.
 
         Args:
@@ -70,7 +80,7 @@ class ApexOptimWrapper(OptimWrapper):
         state_dict['apex_amp'] = apex_amp.state_dict()
         return state_dict
 
-    def load_state_dict(self, state_dict: dict):
+    def load_state_dict(self, state_dict: dict) -> None:
         """Load and parse the state dictionary of :attr:`optimizer` and
         :attr:`apex_amp`.
 
@@ -95,15 +105,16 @@ class ApexOptimWrapper(OptimWrapper):
         Args:
             model (nn.Module): The training model.
         """
-        if hasattr(self.optimizer, '_amp_stash'):
-            yield
-        else:
+        with super().optim_context(model):
+            # when a given optimizer be passed through apex_amp.initialize,
+            # the "_amp_stash" property will be added
+            if hasattr(self.optimizer, '_amp_stash'):
+                yield
             if isinstance(model, torch.nn.parallel.DistributedDataParallel):
                 model = model.module
-            with super().optim_context(model):
-                model, self.optimizer = apex_amp.initialize(
-                    model,
-                    self.optimizer,
-                    opt_level=self.opt_level,
-                    loss_scale=self.loss_scale)
-                yield
+            model, self.optimizer = apex_amp.initialize(
+                model,
+                self.optimizer,
+                opt_level=self.opt_level,
+                loss_scale=self.loss_scale)
+            yield
--- a/mmengine/optim/optimizer/apex_optimizer_wrapper.py
+++ b/mmengine/optim/optimizer/apex_optimizer_wrapper.py
@@ -158,24 +158,22 @@ class ApexOptimWrapper(OptimWrapper):
         with super().optim_context(model):
             # when a given optimizer be passed through apex_amp.initialize,
             # the "_amp_stash" property will be added
-            if hasattr(self.optimizer, '_amp_stash'):
-                yield
-                return
-            if mmengine.model.wrappers.is_model_wrapper(model):
-                model = model.module
-            model, self.optimizer = apex_amp.initialize(
-                model,
-                self.optimizer,
-                opt_level=self.opt_level,
-                loss_scale=self.loss_scale,
-                enabled=self.enabled,
-                cast_model_type=self.cast_model_type,
-                patch_torch_functions=self.patch_torch_functions,
-                keep_batchnorm_fp32=self.keep_batchnorm_fp32,
-                master_weights=self.master_weights,
-                cast_model_outputs=self.cast_model_outputs,
-                num_losses=self.num_losses,
-                verbosity=self.verbosity,
-                min_loss_scale=self.min_loss_scale,
-                max_loss_scale=self.max_loss_scale)
+            if not hasattr(self.optimizer, '_amp_stash'):
+                if mmengine.model.wrappers.is_model_wrapper(model):
+                    model = model.module
+                model, self.optimizer = apex_amp.initialize(
+                    model,
+                    self.optimizer,
+                    opt_level=self.opt_level,
+                    loss_scale=self.loss_scale,
+                    enabled=self.enabled,
+                    cast_model_type=self.cast_model_type,
+                    patch_torch_functions=self.patch_torch_functions,
+                    keep_batchnorm_fp32=self.keep_batchnorm_fp32,
+                    master_weights=self.master_weights,
+                    cast_model_outputs=self.cast_model_outputs,
+                    num_losses=self.num_losses,
+                    verbosity=self.verbosity,
+                    min_loss_scale=self.min_loss_scale,
+                    max_loss_scale=self.max_loss_scale)
             yield
--- a/mmengine/optim/optimizer/apex_optimizer_wrapper.py
+++ b/mmengine/optim/optimizer/apex_optimizer_wrapper.py
@@ -10,7 +10,7 @@ from .optimizer_wrapper import OptimWrapper
 try:
     import apex.amp as apex_amp
 except ImportError:
-    pass
+    apex_amp = None
 
 
 @OPTIM_WRAPPERS.register_module()
@@ -37,6 +37,9 @@ class ApexOptimWrapper(OptimWrapper):
     """
 
     def __init__(self, opt_level='O1', loss_scale='dynamic', **kwargs):
+        assert apex_amp is not None, \
+            'Apex is not installed. Please check ' \
+            'https://github.com/NVIDIA/apex#linux.'
         super().__init__(**kwargs)
         self.opt_level = opt_level
         self.loss_scale = loss_scale
--- a/mmengine/optim/optimizer/apex_optimizer_wrapper.py
+++ b/mmengine/optim/optimizer/apex_optimizer_wrapper.py
@@ -5,6 +5,9 @@ from typing import Optional, Union
 import torch
 import torch.nn as nn
 
+# a circular import will be caused by
+# from mmengine.model.wrappers import is_model_wrapper
+import mmengine
 from mmengine.registry import OPTIM_WRAPPERS
 from .optimizer_wrapper import OptimWrapper
 
@@ -158,7 +161,7 @@ class ApexOptimWrapper(OptimWrapper):
             if hasattr(self.optimizer, '_amp_stash'):
                 yield
                 return
-            if isinstance(model, torch.nn.parallel.DistributedDataParallel):
+            if mmengine.model.wrappers.is_model_wrapper(model):
                 model = model.module
             model, self.optimizer = apex_amp.initialize(
                 model,
--- a/mmengine/optim/optimizer/apex_optimizer_wrapper.py
+++ b/mmengine/optim/optimizer/apex_optimizer_wrapper.py
@@ -23,12 +23,11 @@ class ApexOptimWrapper(OptimWrapper):
     ``OptimWrapper``, so ``ApexOptimWrapper`` can be used in the same way
     as ``OptimWrapper``.
 
-    Warnings:
+    Warning:
         ``ApexOptimWrapper`` requires `nvidia apex
         <https://github.com/NVIDIA/apex>`_
 
     Args:
-
         opt_level (str, default="O1"): Pure or mixed precision
             optimization level. Accepted values are "O0", "O1", "O2",
             and "O3".
--- a/mmengine/optim/optimizer/apex_optimizer_wrapper.py
+++ b/mmengine/optim/optimizer/apex_optimizer_wrapper.py
@@ -1,6 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 from contextlib import contextmanager
-from typing import Union
+from typing import Optional, Union
 
 import torch
 import torch.nn as nn
@@ -35,6 +35,35 @@ class ApexOptimWrapper(OptimWrapper):
         loss_scale (float or str, default=None): If passed as
             a string, must be a string representing a number,
             e.g., "128.0", or the string "dynamic".
+        enabled (bool, default=True): If False, renders all Amp calls no-ops,
+            so your script should run as if Amp were not present.
+        cast_model_type (torch.dtype, default=None): Model's parameters and
+            buffers to the desired type.
+        patch_torch_functions (bool, default=None): Patch all Torch functions
+            and Tensor methods to perform Tensor Core-friendly ops like GEMMs
+            and convolutions in FP16,
+            and any ops that benefit from FP32 precision in FP32.
+        keep_batchnorm_fp32 (bool or str, default=None): To enhance precision
+            and enable cudnn batchnorm (which improves performance),
+            it's often beneficial to keep batchnorm weights in FP32
+            even if the rest of the model is FP16.
+            If passed as a string, must be the string "True" or "False".
+        master_weights (bool, default=None): Maintain FP32 master weights to
+            accompany any FP16 model weights. FP32 master weights are stepped
+            by the optimizer to enhance precision and capture small gradients.
+        cast_model_outputs (torch.dtype, default=None): Option to ensure that
+            the outputs of your model(s) are always cast to a particular type
+            regardless of ``opt_level``.
+        num_losses (int, default=1): Option to tell Amp in advance how many
+            losses/backward passes you plan to use.
+        verbosity (int, default=1): Set to 0 to suppress Amp-related output.
+        min_loss_scale (float, default=None): Sets a floor for the loss scale
+            values that can be chosen by dynamic loss scaling.
+            The default value of None means that no floor is imposed.
+            If dynamic loss scaling is not used, `min_loss_scale` is ignored.
+        max_loss_scale (float, default=2.**24): Sets a ceiling for the
+            loss scale values that can be chosen by dynamic loss scaling.
+            If dynamic loss scaling is not used, `max_loss_scale` is ignored.
         **kwargs: Keyword arguments passed to OptimWrapper.
 
     Note:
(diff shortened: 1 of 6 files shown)


HAOCHENYE (commented):
Hi! Thanks for your contribution. It seems the current implementation does not use `apex.amp.initialize` to prepare the model and optimizer,  I know there exits some limitations that make it hard to be implemented In MMEngine, and I want to discuss how we could support `ApexOptimWrapper` better.

1.If we don't call initialize, will mix-precision training work?
 
2.As the official document of apex describes:

image

We need to call amp.initialize() before wrap the model with ddp wrapper and optimizer. However,  there is a confliction, PyTorch want suggest we given the parameters of ddp model to optimizer in [tutorial](https://pytorch.org/tutorials/intermediate/ddp_tutorial.html?utm_source=distr_landing&utm_medium=intermediate_ddp_tutorial) 

image

Actually, it is reasonable, since distributed wrapper like [FSDP](https://pytorch.org/docs/stable/fsdp.html?highlight=fsdp#module-torch.distributed.fsdp) will overwrite the original parameters and we have to pass its parameters to optimizer after wrapping the model.  Alright, it means that when we got the optimizer, the model must have been wrapped by DDP, which conflicts with the principle of apex.

In MMEngine, `OptimWrapper.optim_context` can get the ddp-model and optimizer, I'm not sure if we can use `amp.initialize` there (Maybe initialize the model in ddp wrapper in place?).

3.We need to consider how to resume the optimizer.

xcnick:
> Hi! Thanks for your contribution. It seems the current implementation does not use `apex.amp.initialize` to prepare the model and optimizer, I know there exits some limitations that make it hard to be implemented In MMEngine, and I want to discuss how we could support `ApexOptimWrapper` better.
> 
> 1.If we don't call initialize, will mix-precision training work?
> 
> 2.As the official document of apex describes:
> 
> image
> 
> We need to call amp.initialize() before wrap the model with ddp wrapper and optimizer. However, there is a confliction, PyTorch want suggest we given the parameters of ddp model to optimizer in [tutorial](https://pytorch.org/tutorials/intermediate/ddp_tutorial.html?utm_source=distr_landing&utm_medium=intermediate_ddp_tutorial)
> 
> image
> 
> Actually, it is reasonable, since distributed wrapper like [FSDP](https://pytorch.org/docs/stable/fsdp.html?highlight=fsdp#module-torch.distributed.fsdp) will overwrite the original parameters and we have to pass its parameters to optimizer after wrapping the model. Alright, it means that when we got the optimizer, the model must have been wrapped by DDP, which conflicts with the principle of apex.
> 
> In MMEngine, `OptimWrapper.optim_context` can get the ddp-model and optimizer, I'm not sure if we can use `amp.initialize` there (Maybe initialize the model in ddp wrapper in place?).
> 
> 3.We need to consider how to resume the optimizer.

1. `apex.amp.initialize` must be called to use mix-precision training.
2. I think (not sure) `apex.amp.initialize` is independent of DDP, so the pipeline is:
```python
my_model = Model()
my_opt = SGD(...)
my_model, my_opt =  apex.amp.initialize(my_model, my_opt, opt_level='O1', loss_scale=...)
ddp_model = DDP(my_model)
```
3. `ApexOptimWrapper` use the same `load_state_dict()` and `state_dict()` method as `OptimWrapper`.

HAOCHENYE:
> 1. `apex.amp.initialize` must be called to use mix-precision training.
> 2. I think (not sure) `apex.amp.initialize` is independent of DDP, so the pipeline is:
> 
> ```python
> my_model = Model()
> my_opt = SGD(...)
> my_model, my_opt =  apex.amp.initialize(my_model, my_opt, opt_level='O1', loss_scale=...)
> ddp_model = DDP(my_model)
> ```
> 
> 3. `ApexOptimWrapper` use the same `load_state_dict()` and `state_dict()` method as `OptimWrapper`.

Ahhh, If we want to use ApexOptimWrapper independent of `Runner`, I think the current implementation is almost enough. But if we want to use `ApexOptimWrapper` in Runner, what should I do?

The key problem is that **where**  should we call `apex.amp.initialize` to enable mix-precision training. It seems `ApexOptimWrapper`  now does not call `apex.amp.initialize`. Does it mean that if we simply replace the `OptimWrapper` with `ApexOptimWrapper`, the mixed-precision training will not be enabled?

The second problem is the compatibility of ddp-training, as mentioned above:

image

`amp.initialize` require that the model should not be wrapper with ddp-wrapper. Therefore, how could we do this in `Runner` ?
 
3. load or resume

In the [apex tutorial](
image
)

image

apex.amp has their own way t save or load checkpoint, we should take it into consideration.

nijkah:
Hi, I also have an interest in this feature. 
Related to https://github.com/open-mmlab/mmengine/issues/627, we have our own implementation to support `DeepSpeed` in `MMEngine` and willingness to post the PR.

One of the crucial changes to support `DeepSpeed` is that `deepspeed` inlcude `deepspeed.initialize(model=model, optimizer=optimizer)` which is a very similar interface as `apex.amp.initialize`.

As mentioned by @HAOCHENYE , these `initialize` interfaces should be called before `Runner.wrap_model` and it is not possible with `OptimWrapper.optim_context`.

So we had to write the new `DeepSpeedRunner` to support it.
I think it is a quite key feature to several related frameworks such as [Colossal-AI](https://colossalai.org/docs/basics/initialize_features/).

Can we discuss this? (New Issue or Discussion?_

xcnick:
@HAOCHENYE 
As you metioned, the key problem is that where should we call apex.amp.initialize to enable mix-precision training.
3.load and resume, also depends on where `apex.amp.initialize` is called. If `ApexOptimWrapper` does not call `apex.amp.initialize`, just keep the existing implementation, inherit from `OptimWrapper`, `amp.state_dict()` will be handled somewhere outside, such as `ApexRunner`.

HAOCHENYE:
@nijkah 
Good suggestions!! Recently we are (mainly explored by @C1rN09) also committed to figuring out a proper design to support training with DeepSpeed or Colossal AI.  I think it could be a good idea to create a new discussion to discuss this topic Specially
 😆 ！

@xcnick 
Completely agree! But could we call `apex.amp.initialize` in some tricky way? Actually, `_initialize` in apex

https://github.com/NVIDIA/apex/blob/082f999a6e18a3d02306e27482cc7486dab71a50/apex/amp/_initialize.py#L145

only patch the `forward` and register some custom hooks into `forward`, I'm not sure it will work when we call `amp.initialize` in `optim_context` like:

```python
# suppose in ddp training
amp.initialize(model.module, self.optimizer)
```

If it works, I think it could be a temporary solution to support `ApexOptimWrapper`. When we find a more elegant way to support Colossal AI or DeepSpeed, which also applies to `ApexOptimWrapper`, we could switch to the new design.

C1rN09:
> Can we discuss this? (in this PR, the new Issue or discussion board)

Hi, @nijkah ! I've created a [discussion thread](https://github.com/open-mmlab/mmengine/discussions/749). You can paste these comments there and we can discuss on it!

And @xcnick since this PR is related, your opinions and ideas are also welcome!

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,106 @@
...
+        **kwargs: Keyword arguments passed to OptimWrapper.
+
+    Note:
+        If you use ``IterBasedRunner`` and enable gradient accumulation,
+        the original `max_iters` should be multiplied by
+        ``accumulative_counts``.
+    """
+
+    def __init__(self, opt_level='O1', loss_scale='dynamic', **kwargs):
+        super().__init__(**kwargs)
```
nijkah, line 40:
```suggestion
    def __init__(self, opt_level='O1', loss_scale='dynamic', **kwargs):
    assert apex_amp is not None, \
        'Apex is not installed. Please check https://github.com/NVIDIA/apex#linux.',
        super().__init__(**kwargs)
```
The assertion logic similar to the line in the test code could be added here.

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,106 @@
...
+import torch
+import torch.nn as nn
+
+from mmengine.registry import OPTIM_WRAPPERS
+from .optimizer_wrapper import OptimWrapper
+
+try:
+    import apex.amp as apex_amp
+except ImportError:
+    pass
```
nijkah, line 13:
```suggestion
except ImportError:
    apex_amp = None
```

xcnick:
@nijkah Thanks for your suggestion.

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,109 @@
...
+            model (nn.Module): The training model.
+        """
+        if hasattr(self.optimizer, '_amp_stash'):
+            yield
+        else:
+            if isinstance(model, torch.nn.parallel.DistributedDataParallel):
+                model = model.module
+            with super().optim_context(model):
+                model, self.optimizer = apex_amp.initialize(
+                    model,
```
nijkah, line 105: This is just a question. When I checked the Apex documentation, I couldn't find a description to handle it like this.
Isn't there any issue in training when handling it like this?

HAOCHENYE: This is actually what I concerned, I took a cursory look at some of implementation for `apex.initialize`:

https://github.com/NVIDIA/apex/blob/082f999a6e18a3d02306e27482cc7486dab71a50/apex/amp/_initialize.py#L145

It seems `initialize` only patches the forward and registers some custom hooks into forward. I'm not sure about doing this on `model.module` will work or not, we need to check the optimization for saving memory and accelerating training when `ApexOptimWrapper` is enabled.

@xcnick could you provide the comparison of `nvidia-smi` when `ApexOptimWrapper` is enabled or not?

xcnick: OK, there is something wrong with current implementation.

In `examples\examples/distributed_training.py`, compare the following two options:
```python
batch_size=1024,
...
optim_wrapper=dict(
        type='ApexOptimWrapper', opt_level='O1', loss_scale=10,
        optimizer=dict(type=Adam, lr=0.001)),
...
```
and

```python
batch_size=1024,
...
optim_wrapper=dict(optimizer=dict(type=SGD, lr=0.001, momentum=0.9)),
```

result:

| optim_wrapper | nvidia-smi | memory in log file |
| -- | -- | -- |
| ApexOptimWrapper |  11GB | 5574 |
| original OptimWrapper | 5.2GB | 2249 |

It seems that calling `amp.initialize` in `optim_context` may not be the correct way.

nijkah: You may be able to participate in the related discussion if you are interested.
https://github.com/open-mmlab/mmengine/discussions/749#discussioncomment-4201408

HAOCHENYE: Emmm, the conclusion makes me confused. I've tried to train mmdet with retinanet-r50 based on this PR.  The result of nvidia-smi is:

image

image

It seems that the `ApexOptimWrapper` has worked.... I also test the `examples/examples/distributed_training.py`, the result is the same.

So... what happened 🤣

xcnick: That is so embarrassing.

In addition, I've tested `O0` / `O2` / `O3`:
- `O0`: memory usage is almost the same as the `OptimWrapper`
- `O3`: memory usage is between `O0` and `O1`, and the loss is nan all the time
- `O2`: some compatibility errors occur in the runner

Also, some compatibility modifications are required when apex amp is set to `O0` or `O3`.

HAOCHENYE: I'll use `ApexOptimWrapper` based on this PR to train MMDet these days, and check the training speed and accuracy. Do you have any suggestions or comments about the current implementations @xcnick @nijkah?

xcnick: I'm not sure whether the this implementation can work correctly, so good luck ^^

In addition, the `to` method of `BaseModel` in `base_model.py` needs to be modified for compatibility in `O0` and `O3` modes.

C1rN09: > In addition, the `to` method of `BaseModel` in `base_model.py` needs to be modified for compatibility in `O0` and `O3` modes.

Do you mean the modifications in #783 ? If it is not enough, feel free to leave a comment!

xcnick: Great job!

HAOCHENYE: Sorry for my late reply. To give a conclusion first: the current implementation seems OK. I used `ApexOptimWrapper` to train the `atss` in MMDet, and the loss was able to converge normally, and it will take some time to verify the accuracy.

The reason for the late reply is that I spent some time trying to train RetinaNet (O1) based on `ApexOptimWrapper`. However, it turned out that neither `AmpOptimWrapper` nor `ApexOptimWrapper· can train RetinaNet (loss is `nan`) normally, which is determined by the model's own characteristics and has nothing to do with the implementation of `ApexOptimWrapper`

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,109 @@
...
+    def optim_context(self, model: nn.Module):
+        """Enables the context for mixed precision training, and enables the
+        context for disabling gradient synchronization during gradient
+        accumulation context.
+
+        Args:
+            model (nn.Module): The training model.
+        """
+        if hasattr(self.optimizer, '_amp_stash'):
+            yield
```
HAOCHENYE, line 99:
```suggestion
        if hasattr(self.optimizer, '_amp_stash'):
            with super().optim_context(model):
                yield
```

HAOCHENYE: Does it mean that the initialized optimizer will have the `_amp_stash` attribute? Maybe we need to add some comments here.

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,109 @@
...
+    """A subclass of :class:`OptimWrapper` that supports automatic mixed
+    precision training based on apex.amp.
+
+    ``ApexOptimWrapper`` provides a unified interface with
+    ``OptimWrapper``, so ``ApexOptimWrapper`` can be used in the same way
+    as ``OptimWrapper``.
+
+    Warnings:
+        ``ApexOptimWrapper`` requires
+            [nvidia apex](https://github.com/NVIDIA/apex).
```
HAOCHENYE, line 27:
```suggestion
        ``ApexOptimWrapper`` requires `nvidia apex <https://github.com/NVIDIA/apex>`_
```

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,109 @@
...
+
+    Args:
+
+        **kwargs: Keyword arguments passed to OptimWrapper.
```
HAOCHENYE, line 31: Missing arguments description here

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,109 @@
...
+
+    Note:
+        If you use ``IterBasedRunner`` and enable gradient accumulation,
+        the original `max_iters` should be multiplied by
+        ``accumulative_counts``.
+    """
+
+    def __init__(self, opt_level='O1', loss_scale='dynamic', **kwargs):
```
HAOCHENYE, line 39: missing type hint

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,109 @@
...
+        assert apex_amp is not None, \
+            'Apex is not installed. Please check ' \
+            'https://github.com/NVIDIA/apex#linux.'
+        super().__init__(**kwargs)
+        self.opt_level = opt_level
+        self.loss_scale = loss_scale
+
+    def backward(self, loss: torch.Tensor, **kwargs):
```
HAOCHENYE, line 47:
```suggestion
    def backward(self, loss: torch.Tensor, **kwargs) -> None:
```

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,109 @@
...
+
+        Returns:
+            dict: The merged state dict of :attr:`apex_amp` and
+            :attr:`optimizer`.
+        """
+        state_dict = self.optimizer.state_dict()
+        state_dict['apex_amp'] = apex_amp.state_dict()
+        return state_dict
+
+    def load_state_dict(self, state_dict: dict):
```
HAOCHENYE, line 73:
```suggestion
    def load_state_dict(self, state_dict: dict) -> None:
```

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,109 @@
...
+
+        Args:
+            model (nn.Module): The training model.
+        """
+        if hasattr(self.optimizer, '_amp_stash'):
+            yield
+        else:
+            if isinstance(model, torch.nn.parallel.DistributedDataParallel):
+                model = model.module
+            with super().optim_context(model):
```
HAOCHENYE, line 103: `super().optim_context` should be called each iteration for avoiding necessary gradient accumulation during training,

HAOCHENYE (commented):
Hi~ is there any progress? Besides, CLA should be signed again :rofl: .

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,120 @@
...
+            model (nn.Module): The training model.
+        """
+        with super().optim_context(model):
+            # when a given optimizer be passed through apex_amp.initialize,
+            # the "_amp_stash" property will be added
+            if hasattr(self.optimizer, '_amp_stash'):
+                yield
+            if isinstance(model, torch.nn.parallel.DistributedDataParallel):
+                model = model.module
+            model, self.optimizer = apex_amp.initialize(
```
zhouzaida, line 115: Is it necessary to support passing in other parameters? https://nvidia.github.io/apex/amp.html#apex.amp.initialize

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def backward(self, loss: torch.Tensor, **kwargs) -> None:
+        """Perform gradient back propagation with :attr:`loss_scaler`.
+
+        Args:
+            loss (torch.Tensor): The loss of current iteration.
+            kwargs: Keyword arguments passed to :meth:`torch.Tensor.backward`
+        """
+        with apex_amp.scale_loss(loss, self.optimizer) as scaled_loss:
+            scaled_loss.backward()
```
zhouzaida, line 65: Can it accept `**kwargs`?

xcnick: Of course, `scaled_loss` is a `torch.Tensor`.

zhouzaida:
Please add `ApexOptimWrapper` in https://github.com/open-mmlab/mmengine/blob/main/docs/en/api/optim.rst#optimizer and `https://github.com/open-mmlab/mmengine/blob/main/docs/zh_cn/api/optim.rst#optimizer`

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,179 @@
...
+@OPTIM_WRAPPERS.register_module()
+class ApexOptimWrapper(OptimWrapper):
+    """A subclass of :class:`OptimWrapper` that supports automatic mixed
+    precision training based on apex.amp.
+
+    ``ApexOptimWrapper`` provides a unified interface with
+    ``OptimWrapper``, so ``ApexOptimWrapper`` can be used in the same way
+    as ``OptimWrapper``.
+
+    Warnings:
```
zhouzaida, line 26:
```suggestion
    Warning:
```

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,179 @@
...
+        ``ApexOptimWrapper`` requires `nvidia apex
+        <https://github.com/NVIDIA/apex>`_
+
+    Args:
+
```
zhouzaida, line 31:
```suggestion
```

xcnick:
> Please add `ApexOptimWrapper` in https://github.com/open-mmlab/mmengine/blob/main/docs/en/api/optim.rst#optimizer and `https://github.com/open-mmlab/mmengine/blob/main/docs/zh_cn/api/optim.rst#optimizer`

done.

zhouzaida:
Hi @xcnick , thanks for your contribution. You could pull the main branch to fix the lint error caused by `isort`. More details at #890

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,120 @@
...
+        accumulation context.
+
+        Args:
+            model (nn.Module): The training model.
+        """
+        with super().optim_context(model):
+            # when a given optimizer be passed through apex_amp.initialize,
+            # the "_amp_stash" property will be added
+            if hasattr(self.optimizer, '_amp_stash'):
+                yield
```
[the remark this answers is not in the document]
xcnick: It's weird that `yield` is followed by `return`, but it make sense in contextmanager.

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,178 @@
...
+                return
+            if isinstance(model, torch.nn.parallel.DistributedDataParallel):
```
zhouzaida, line 161: Should we use `is_model_wrapper` to check this condition?

https://github.com/open-mmlab/mmengine/blob/6dc1d7082a1e2bc32049e65f283c4fd20de1c026/mmengine/model/wrappers/utils.py#L7

tests/test_optim/test_optimizer/test_optimizer_wrapper.py
```diff
@@ -283,6 +290,100 @@ def mock_methd(loss):
         optim_wrapper.zero_grad = MagicMock()
 
 
+class TestApexOptimWrapper(TestCase):
```
zhouzaida, line 293:
```suggestion
@unittest.skipIf(not torch.cuda.is_available(), reason='need gpu to test Apex')
class TestApexOptimWrapper(TestCase):
```

zhouzaida:
Hi @xcnick , thanks for your contribution. This PR can be merged after resolving the above final comments and validating it in your local machine.

zhouzaida (dismissed):

HAOCHENYE:
Hi! as the official document says:

image

The current implementation of `ApexOptimWrapper.load_state_dict` will raise an error for the lack of initialization of `apex`. A possibly workaround could be not loading `state_dict` in `ApexOptimWrapper.load_state_dict`, but only saving the `state_dict` as an attribute, then calling `apex.amp.load_state_dict` in  `optim_context`

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -176,4 +183,8 @@ def optim_context(self, model: nn.Module):
                     verbosity=self.verbosity,
                     min_loss_scale=self.min_loss_scale,
                     max_loss_scale=self.max_loss_scale)
+                # loading apex_amp state_dict after initialization of apex_amp
+                if self._apex_amp_state_dict:
```
HAOCHENYE, line 187:
```suggestion
                if self._apex_amp_state_dict is not None:
```

mmengine/optim/optimizer/apex_optimizer_wrapper.py
```diff
@@ -0,0 +1,190 @@
...
+            `apex_amp.initialize` is called.
+        Args:
+            state_dict (dict): The state dict of :attr:`optimizer` and
+                :attr:`apex_amp`
+        """
+        if 'apex_amp' in state_dict:
+            if hasattr(self.optimizer, '_amp_stash'):
+                apex_amp.load_state_dict(state_dict.pop('apex_amp'))
+            else:
+                self._apex_amp_state_dict = state_dict.pop('apex_amp')
```
zhouzaida, line 153: It would be better to explain why we temporarily cache the `apex_amp` rather than directly load it to `apex_amp` object.

xcnick: ok, I added some comments.

zhouzaida (approved):

HAOCHENYE (approved):

PR #783, opened 2022-12-02
[Fix] BaseModel & BaseDataPreprocessor `to` method to be consistent with torch.nn.Module

C1rN09:

## Motivation

Make `BaseModel.to` and `BaseDataPreprocessor.to` methods to be consistent with `torch.nn.Module.to`, i.e. it can accept devices and dtypes as arguments and handle these cases properly.

Before this PR:

```python
class Model(BaseModel):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(4, 4)
    def forward(self):
        pass

model = Model()
model.to('cuda')  # correct
model.to(torch.half)  # raise TypeError
model.to(device='cuda', dtype=torch.half)  # no exception, but dtype ignored
```

After this PR:

```python
model.to('cuda')
model.to(torch.half)
model.to('cpu', torch.float64)
model.to('cpu', dtype=torch.float64)
model.to(torch.half, device='cpu')  # error, this is not supported by nn.Module
```

## Modification

As in PR

## BC-breaking (Optional)

Probably, because before this PR `dtype` kwargs are ignored

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/model/base_model/base_model.py
+++ b/mmengine/model/base_model/base_model.py
@@ -155,9 +155,9 @@ class BaseModel(BaseModule):
 
         Returns:
             tuple[Tensor, dict]: There are two elements. The first is the
-                loss tensor passed to optim_wrapper which may be a weighted sum
-                of all losses, and the second is log_vars which will be sent to
-                the logger.
+            loss tensor passed to optim_wrapper which may be a weighted sum
+            of all losses, and the second is log_vars which will be sent to
+            the logger.
         """
         log_vars = []
         for loss_name, loss_value in losses.items():
--- a/mmengine/model/base_model/base_model.py
+++ b/mmengine/model/base_model/base_model.py
@@ -184,15 +184,7 @@ class BaseModel(BaseModule):
         Returns:
             nn.Module: The model itself.
         """
-        if 'device' in kwargs:
-            device = kwargs['device']
-        elif args:
-            try:
-                device = torch.device(args[0])
-            except TypeError:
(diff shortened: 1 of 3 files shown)


HAOCHENYE (dismissed):
Approved!!!
image

mmengine/model/base_model/base_model.py
```diff
@@ -177,23 +177,17 @@ def parse_losses(
...
         additionally.
 
-        Args:
-            device (int, str or torch.device, optional): the desired device
-                of the parameters and buffers in this module.
-
         Returns:
             nn.Module: The model itself.
         """
+        device = torch._C._nn._parse_to(*args, **kwargs)[0]
```
zhouzaida, line 187: Using an internal function may cause a breaking issue.

C1rN09: I've tried writing a parsing method myself, but finally I believe this internal function is more stable. [Lightning](https://github.com/Lightning-AI/lightning/blob/master/src/lightning_lite/utilities/device_dtype_mixin.py#L52) also uses it directly.

zhouzaida: What is the minimum pytorch version that provides this interface, and is it greater than or equal to 1.6?

C1rN09: I checked pytorch repo with their tags. It appears since v1.0.0

zhouzaida (approved):

PR #795, opened 2022-12-07
[Fix] Add preprocess inputs to inference

HAOCHENYE:

## Motivation

`inferencer.visualize` needs the original inputs rather than the preprocessed inputs to show the visualization result, therefore we need to pass the original inputs to `visualize`. 

This PR extract some logics from `preprocess` to `preprocess_data`, which is responsible for listing the files from a directory and converting the inputs to a list. `__call__` will call preprocess_data and pass the returned value to `visualize`

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/infer/infer.py
+++ b/mmengine/infer/infer.py
@@ -203,7 +203,7 @@ class BaseInferencer(metaclass=InferencerMeta):
             postprocess_kwargs,
         ) = self._dispatch_kwargs(**kwargs)
 
-        ori_inputs = self.preprocess_inputs(inputs)
+        ori_inputs = self._inputs_to_list(inputs)
         inputs = self.preprocess(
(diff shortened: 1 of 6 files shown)


mmengine/infer/infer.py
```diff
@@ -203,17 +203,51 @@ def __call__(
...
 
+    def preprocess_inputs(self, inputs: InputsType) -> list:
+        """Preprocess the inputs to a list.
+
+        Preprocess inputs to a list according to its type:
+
+            - list or tuple: return inputs
+            - str:
+                - Directory path: return all files in the directory
+                - normal string: return a list containing the string
```
zhouzaida, line 227: What is the case of normal string?

PR #808, opened 2022-12-10
[Refactor] Refactor vis hook

HAOCHENYE:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


tests/data/config/py_config/test_custom_class.py
zhouzaida, line 1: This file should not be updated in this PR.

HAOCHENYE: As the PR title says, this PR will refactor the implementation of `NaiveVisualizationHook`. Here are some reasons why I will do this:

1. None of the downstream repo uses the `NaiveVisualizationHook` 
2. `NaiveVisualizationHook` should not assume that the result processed by dataset contain the field like `img_norm_cfg` and `pad_shape`. The result fetched from dataset  should only meet the minimum requirement, for example, contain the key `data_sample` and `inputs`, which is consistent with the assumption of `ImgDataPreprocessor`

PR #817, opened 2022-12-12
[Docs] Translate registry.md

xin-li-67:
## Motivation

translate docs/en/advanced_tutorials/registry.md 

## Modification

docs/en/advanced_tutorials/registry.md 

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


docs/en/advanced_tutorials/registry.md
```diff
@@ -1,3 +1,302 @@
...
+
+## What is a registry
+
+The [registry](mmengine.registry.Registry) implemented in MMEngine can be considered as a combination of mapping tables and module build functions. The mapping table maintains a mapping of strings to **classes or functions**, allowing the user to find the corresponding class or function with the help of a string. For example, the mapping of the string `"ResNet"` to the `ResNet` class or function allows users to find the `ResNet` class with the string `"ResNet"`. The module build function defines how to find the corresponding class or function based on a string and how to instantiate the class or call the function.For example, finding `nn.BatchNorm2d` and instantiating the `BatchNorm2d` module by the string `"bn"`, or finding the `build_batchnorm2d` function by the string `"build_batchnorm2d"` and then returning the result. The registries in MMEngine use the [build_from_cfg](mmengine.registry.build_from_cfg) function by default to find and instantiate the class or function corresponding to the string.
```
C1rN09, line 7:
```suggestion
The [registry](mmengine.registry.Registry) in MMEngine can be considered as a union of a mapping table and a build function of modules. The mapping table maintains a mapping from strings to **classes or functions**, allowing the user to find the corresponding class or function with its name/notation. For example, the mapping from the string `"ResNet"` to the `ResNet` class. The module build function defines how to find the corresponding class or function based on a string and how to instantiate the class or call the function. For example, finding `nn.BatchNorm2d` and instantiating the `BatchNorm2d` module by the string `"bn"`, or finding the `build_batchnorm2d` function by the string `"build_batchnorm2d"` and then returning the result. The registries in MMEngine use the [build_from_cfg](mmengine.registry.build_from_cfg) function by default to find and instantiate the class or function corresponding to the string.
```

docs/en/advanced_tutorials/registry.md
```diff
@@ -1,3 +1,302 @@
...
+def print_args(**kwargs):
+    print(kwargs)
+
+func_cfg = dict(type='print_args', a=1, b=2)
+func_res = FUNCTION.build(func_cfg)
+```
+
+## Advanced usage
+
+The registry in MMEngine supports hierarchical registration, which enables cross-project calls, meaning that modules from another project can be used in another project. Though there are other ways to implement this, the registry provides a much easier solution.
```
C1rN09, line 161:
```suggestion
The registry in MMEngine supports hierarchical registration, which enables cross-project calls, meaning that modules from one project can be used in another project. Though there are other ways to implement this, the registry provides a much easier solution.
```

docs/en/advanced_tutorials/registry.md
```diff
@@ -1,3 +1,302 @@
...
+print(ACTIVATION.module_dict)
+# {
+#     'Sigmoid': __main__.Sigmoid,
+#     'ReLU': __main__.ReLU,
+#     'Softmax': __main__.Softmax
+# }
+```
+
+```{note}
+The registry mechanism will only be triggered when the corresponded module file is imported, so we need to import the file somewhere or dynamically import the module using the ``custom_imports`` field to trigger the mechanism. Please refer to [Importing custom Python modules](config.md) for more details.
```
C1rN09, line 78:
```suggestion
The registry mechanism will only be triggered when the corresponded module file is imported, so we need to import the file somewhere or dynamically import the module using the ``custom_imports`` field to trigger the mechanism. Please refer to [Importing custom Python modules](config.md#import-the-custom-module) for more details.
```

C1rN09: 中文部分也需要改一下，小标题里面的大写字母变小写字母，然后空格变成 - 即可

docs/en/advanced_tutorials/registry.md
```diff
@@ -1,3 +1,302 @@
 # Registry
 
-Coming soon. Please refer to [chinese documentation](https://mmengine.readthedocs.io/zh_CN/latest/advanced_tutorials/registry.html).
+OpenMMLab supports a rich collection of algorithms and datasets, therefore, many modules with similar functionality are implemented. For example, the implementations of `ResNet` and `SE-ResNet` are based on the classes `ResNet` and `SEResNet`, respectively, which have similar functions and interfaces and belong to the model components of the algorithm library. To manage these functionally similar modules, MMEngine implements the [registry](mmengine.registry.registry). Most of the algorithm libraries in OpenMMLab use registrars to manage their code modules, including [MMDetection](https://github.com/open-mmlab/mmdetection), [MMDetection3D](https://github.com/open-mmlab/mmdetection3d), [MMClassification](https://github.com/open-mmlab/mmclassification) and [MMEditing](https://github.com/open-mmlab/mmediting) etc.
```
zhouzaida, line 3:
```suggestion
OpenMMLab supports a rich collection of algorithms and datasets, therefore, many modules with similar functionality are implemented. For example, the implementations of `ResNet` and `SE-ResNet` are based on the classes `ResNet` and `SEResNet`, respectively, which have similar functions and interfaces and belong to the model components of the algorithm library. To manage these functionally similar modules, MMEngine implements the [registry](mmengine.registry.registry). Most of the algorithm libraries in OpenMMLab use `registry` to manage their modules, including [MMDetection](https://github.com/open-mmlab/mmdetection), [MMDetection3D](https://github.com/open-mmlab/mmdetection3d), [MMClassification](https://github.com/open-mmlab/mmclassification) and [MMEditing](https://github.com/open-mmlab/mmediting), etc.
```

docs/en/advanced_tutorials/registry.md
```diff
@@ -1,3 +1,302 @@
...
+
+We can also use the modules of the parent node `MMEngine` here in the `MMAlpha`.
+
+```python
+model = MODELS.build(cfg=dict(type='RReLU', lower=0.2))
+# scope is optional
+model = MODELS.build(cfg=dict(type='mmengine.RReLU'))
+```
+
+If no prefix is added, the `build` method will first find out if the module exists in the current node and return it if there is one. Otherwise it will continue to look up the parent node or even the ancestor node until it finds the module. If the same module exists in both the current node and the parent node, we need to specify the `scope` prefix to indicate that we want to use the module of the parent node.
```
zhouzaida, line 253:
```suggestion
If no prefix is added, the `build` method will first find out if the module exists in the current node and return it if there is one. Otherwise, it will continue to look up the parent nodes or even the ancestor node until it finds the module. If the same module exists in both the current node and the parent nodes, we need to specify the `scope` prefix to indicate that we want to use the module of the parent nodes.
```

docs/en/advanced_tutorials/registry.md
```diff
@@ -1,3 +1,302 @@
...
+
+input = torch.randn(2)
+output = model(input)
+# call RReLU.forward
+print(output)
+```
+
+### Use the module of a sibling node
+
+In addition to using the module of the parent node, users can also call the module of a sibling node.
```
zhouzaida, line 266:
```suggestion
In addition to using the module of the parent nodes, users can also call the module of a sibling node.
```

C1rN09 (approved):
LGTM!

PR #848, opened 2022-12-27
support np.ScalarType in message_hub

ZLTJohn:
## Motivation
After computing metrics, the evaluation result dict will be parsed to mmengine.
However, if the dict item value is a numpy scalar, the message hub will treat it is torch.Tensor in function  `_get_valid_value`.
This would cause error triggered by `assert hasattr(value, 'numel') and value.numel() == 1`.

We should consider other numpy scalar types.

### DIFF
--- a/mmengine/logging/message_hub.py
+++ b/mmengine/logging/message_hub.py
@@ -329,7 +329,7 @@ class MessageHub(ManagerMixin):
             value = value.item()
         elif isinstance(value, (int, float)):
             value = value
-        elif isinstance(value, np.ScalarType):
+        elif isinstance(value, np.number):
             value = value.item()
         else:
             # check whether value is torch.Tensor but don't want
--- a/mmengine/logging/message_hub.py
+++ b/mmengine/logging/message_hub.py
@@ -324,7 +324,7 @@ class MessageHub(ManagerMixin):
         Returns:
             float or int: python built-in type value.
         """
-        if isinstance(value, (np.ndarray,np.number)):
+        if isinstance(value, (np.ndarray, np.number)):
             assert value.size == 1
             value = value.item()
         elif isinstance(value, (int, float)):
--- a/mmengine/logging/message_hub.py
+++ b/mmengine/logging/message_hub.py
@@ -324,13 +324,11 @@ class MessageHub(ManagerMixin):
         Returns:
             float or int: python built-in type value.
         """
-        if isinstance(value, np.ndarray):
+        if isinstance(value, (np.ndarray,np.number)):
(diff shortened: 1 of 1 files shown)


zhouzaida:
> [CLA assistant check](https://cla-assistant.io/open-mmlab/mmengine?pullRequest=848) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/open-mmlab/mmengine?pullRequest=848) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/open-mmlab/mmengine?pullRequest=848) it.

Hi, thanks for your contribution. Please sign the CLA before we merge this PR.

HAOCHENYE:
Hi, it seems `int`, `float`, `str` also belong to `np.ScalarType`

image

But this builtin type does not have the `item()` method. What is the real type of `value` when it triggered an error? Could we have a more accurate check?

HAOCHENYE:
Hi~ Is there any progress?

ZLTJohn:
> Hi, it seems `int`, `float`, `str` also belong to `np.ScalarType`
> 
> image
> 
> But this builtin type does not have the `item()` method. What is the real type of `value` when it triggered an error? Could we have a more accurate check?

Yep, the error types are scalar types that only numpy posses, like np.float32, np.int64 and so on.

HAOCHENYE:
> > Hi, it seems `int`, `float`, `str` also belong to `np.ScalarType`
> > image
> > But this builtin type does not have the `item()` method. What is the real type of `value` when it triggered an error? Could we have a more accurate check?
> 
> Yep, the error types are numeric scalar types that only numpy posses, like np.float32, np.int64 and so on. Those are still numeric types we need to support. However, the existing code fails to support them, since they are NOT np.ndarray.

How about using `isinstance(value, np.number)`?

ZLTJohn:
Great! I use it now.

ZLTJohn:
Sure, it looks great now.

HAOCHENYE:
Hi, the main branch has fixed the CI, please update the code with the latest main.

HAOCHENYE (approved):

zhouzaida:
> Hi, thanks for your contribution. Please sign the CLA before we merge this PR.

Hi @ZLTJohn , thanks for your contribution. Please sign the CLA before we merge this PR.

zhouzaida:
It would be better to add unit tests to cover this modification.

HAOCHENYE:
Hi! CLA shoud be assigned 😄

PR #864, opened 2023-01-05
[Enhance] Support load config from checkpoint.

HAOCHENYE:

## Motivation

Since MMEngine will save the config in the checkpoint, inferencer could be able to load the config from the checkpoint and inference without giving a config.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/infer/infer.py
+++ b/mmengine/infer/infer.py
@@ -167,7 +167,7 @@ class BaseInferencer(metaclass=InferencerMeta):
                     'the config needs to be loaded from the weights')
             cfg = ConfigDict()
         else:
-            raise TypeError('config must be a filepath or any ConfigType'
+            raise TypeError('model must be a filepath or any ConfigType'
                             f'object, but got {type(model)}')
 
         if device is None:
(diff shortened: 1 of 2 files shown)


mmengine/infer/infer.py
```diff
@@ -156,22 +157,23 @@ def __init__(self,
             cfg = copy.deepcopy(model)
         elif isinstance(model, dict):
             cfg = copy.deepcopy(ConfigDict(model))
+        elif model is None:
```
gaotongxiao, line 160: Need to illustrate this feature in docstring

gaotongxiao (approved):

tests/test_infer/test_infer.py
```diff
@@ -107,6 +107,14 @@ def test_init(self):
                 'https://download.openmmlab.com/mmdetection/v2.0/resnest/faster_rcnn_s50_fpn_syncbn-backbone%2Bhead_mstrain-range_1x_coco/faster_rcnn_s50_fpn_syncbn-backbone%2Bhead_mstrain-range_1x_coco_20200926_125502-20289c16.pth',  # noqa: E501
             )
 
+        if is_installed('mmyolo'):
```
zhouzaida, line 110: Should it install mmyolo in CI?

HAOCHENYE: This is only for local test, otherwise, it will influence the statistics of the frequency of downloading the model weight.

mmengine/infer/infer.py
```diff
@@ -156,22 +159,23 @@ def __init__(self,
...
+            assert weights is not None, (
+                'If model is None, the configuration of model must be read '
+                'from weights')
```
zhouzaida, line 165:
```suggestion
                'If model is None, the weights must be specified and contain a config string because the configuration of model must be read '
                'from it')
```

tests/test_infer/test_infer.py
```diff
@@ -107,6 +107,14 @@ def test_init(self):
...
+            from mmyolo.utils import register_all_modules
+            register_all_modules()
+            inferencer = ToyInferencer(
```
hhaAndroid, line 113: A simple way is to temporarily generate a weight in the test, which will be better.  No need to rely on mmyolo

zhouzaida (approved):

PR #879, opened 2023-01-14
To avoid tensor contiguity problems in distributed training

XWHtorrentx:
## Motivation

I had a problem in training mmocr1x model SATRN on a machine with two cuda gpus. It shows that Tensors must be contiguous. 
 
## Modification

I solved it by broadcasting contiguous tensor.

### DIFF
--- a/mmengine/registry/build_functions.py
+++ b/mmengine/registry/build_functions.py
@@ -184,7 +184,7 @@ def build_runner_from_cfg(cfg: Union[dict, ConfigDict, Config],
                     f'{obj_type} is not in the {registry.name} registry. '
                     f'Please check whether the value of `{obj_type}` is '
                     'correct or it was registered as expected. More details '
-                    'can be found at https://mmengine.readthedocs.io/en/latest/tutorials/config.html#import-custom-python-modules'  # noqa: E501
+                    'can be found at https://mmengine.readthedocs.io/en/latest/advanced_tutorials/config.html#import-the-custom-module'  # noqa: E501
                 )
         elif inspect.isclass(obj_type):
             runner_cls = obj_type
@@ -292,7 +292,7 @@ def build_scheduler_from_cfg(
                         f'{scheduler_type} is not in the {registry.name} '
                         'registry. Please check whether the value of '
                         f'`{scheduler_type}` is correct or it was registered '
-                        'as expected. More details can be found at https://mmengine.readthedocs.io/en/latest/tutorials/config.html#import-custom-python-modules'  # noqa: E501
+                        'as expected. More details can be found at https://mmengine.readthedocs.io/en/latest/advanced_tutorials/config.html#import-the-custom-module'  # noqa: E501
                     )
             elif inspect.isclass(scheduler_type):
                 scheduler_cls = scheduler_type
--- a/mmengine/registry/root.py
+++ b/mmengine/registry/root.py
@@ -3,7 +3,7 @@
 projects.
 
 More datails can be found at
-https://mmengine.readthedocs.io/en/latest/tutorials/registry.html.
+https://mmengine.readthedocs.io/en/latest/advanced_tutorials/registry.html.
 """
 
 from .build_functions import (build_model_from_cfg, build_runner_from_cfg,


zhouzaida:
Hi, could you resolve this issue in MMOCR rather than MMEngine?

XWHtorrentx:
> Hi, could you resolve this issue in MMOCR rather than MMEngine?

In fact I dont think there is any problem with MMOCR because I could train this model on single gpu. The traceback info pointed out mmengine trying to broadcast a tensor without checking it is contiguous.

mmengine/dist/dist.py
```diff
@@ -308,7 +308,7 @@ def broadcast(data: Tensor,
         backend_device = get_comm_device(group)
         data_on_device = cast_data_device(data, backend_device)
```
C1rN09, line 309:
```suggestion
        data_on_device: Tensor = cast_data_device(data, backend_device)
```
This should make `mypy` happy

mmengine/dist/dist.py
```diff
@@ -308,7 +308,7 @@ def broadcast(data: Tensor,
...
 
-        torch_dist.broadcast(data_on_device, src, group)
+        torch_dist.broadcast(data_on_device.contiguous(), src, group)
```
C1rN09, line 311:
```suggestion
        data_on_device = data_on_device.contiguous()
        torch_dist.broadcast(data_on_device, src, group)
```
`.contiguous` will produce a temporary variable when the tensor is not contiguous. However, broadcasting a temporary variable is of no use. Should first make `data_on_device` contiguous and then broadcast itself.

C1rN09 (commented):
Can you provide the stacktrace of your error message, so that we can inspect where in MMOCR triggered this error? Currently `mmengine.dist.broadcast` is just a wrapper around `torch.distributed.broadcast`, so I think it's reasonable that it throws error when the input tensor is not contiguous, as long as `torch` does this. What do you think about it? @zhouzaida @HAOCHENYE @Harold-lkk

HAOCHENYE:
Since MMEngine provides a more friendly dis API based PyTorch, for example, users do not need to check `world_size` in broadcast, and do not need to additionally specify `group_list` in gather,   it makes sense to me to omit `contiguous()` operator for users.

XWHtorrentx:
> Can you provide the stacktrace of your error message, so that we can inspect where in MMOCR triggered this error? Currently `mmengine.dist.broadcast` is just a wrapper around `torch.distributed.broadcast`, so I think it's reasonable that it throws error when the input tensor is not contiguous, as long as `torch` does this. What do you think about it? @zhouzaida @HAOCHENYE @Harold-lkk

Traceback (most recent call last):
  File "mmocr/tools/train.py", line 117, in <module>
    main()
  File "mmocr/tools/train.py", line 113, in main
    runner.train()
  File "mmengine/mmengine/runner/runner.py", line 1668, in train
    self._init_model_weights()
  File "mmengine/mmengine/runner/runner.py", line 889, in _init_model_weights
    broadcast(params)
  File "mmengine/mmengine/dist/dist.py", line 311, in broadcast
    torch_dist.broadcast(data_on_device, src, group)
  File "lib/python3.10/site-packages/torch/distributed/distributed_c10d.py", line 1400, in broadcast
    work = default_pg.broadcast([tensor], opts)
RuntimeError: Tensors must be contiguous

mmengine/dist/dist.py
XWHtorrentx: Actually passing a “temporary variable” worked. I have tested this workaround.

C1rN09:
Seems like [these lines in MMOCR](https://github.com/open-mmlab/mmocr/blob/27b6a68586b9a040678fe083bcf60662ae1b9261/mmocr/models/textrecog/layers/satrn_layers.py#L172-L181) created buffers that are not contiguous. I suggest manually making them contiguous @Harold-lkk However, I think this PR also make sense to me because it will make this API more easy to use.

mmengine/dist/dist.py
XWHtorrentx: Thanks for your example. I will modify my PR

C1rN09:
Hi, @XWHtorrentx Thanks for your contribution! The lint failed, can you fix the lint? Besides, it would be better to add some unit tests for this case :smile:

C1rN09 (approved):

HAOCHENYE (approved):

mmengine/dist/dist.py
```diff
@@ -307,7 +307,8 @@ def broadcast(data: Tensor,
         input_device = get_data_device(data)
         backend_device = get_comm_device(group)
         data_on_device = cast_data_device(data, backend_device)
-
+        if isinstance(data_on_device, torch.Tensor):
```
HAOCHENYE, line 310: Is this check necessary? It seems `broadcast` only accepts tensor type inputs

C1rN09: This check is added to pass `mypy` check. Is there any other better solutions?

HAOCHENYE: Will `data_on_device: Tensor = cast_data_device(data, backend_device)` or `assert isinstance(data_on_device, torch.Tensor)` work? If not, I think we should use `# type: ignore` or `if xxx; else raise TypeError` to avoid this confuse `if-else`

XWHtorrentx: This check fixes lint because it passes some other types like Map or Iter. If you have better ways to avoid confusion then please replace this.

zhouzaida:
Hi @XWHtorrentx , I am sorry that this PR was closed for my operation wanted to update the branch to fix the lint. Could you open a new PR to fix this issue?

XWHtorrentx:
> Hi @XWHtorrentx , I am sorry that this PR was closed for my operation wanted to update the branch to fix the lint. Could you open a new PR to fix this issue?

Sure.

PR #910, opened 2023-02-07
[Speed Up] `BaseDataElement` supports `pin_memory` 

hhaAndroid:

## Motivation

Custom object `BaseDataElement` supports `pin_memory` function to speed up training. 

Note: The underlying logic has been modified, it's best to check if the relevant repo has been speeded up.

## Modification

- BaseDataElement add `pin_memory` function
- Add UT

## BC-breaking (Optional)
None

## Use cases (Optional)

dataloader needs to be enabled `pin_memory=True`.

### DIFF
--- a/mmengine/structures/base_data_element.py
+++ b/mmengine/structures/base_data_element.py
@@ -5,6 +5,7 @@ from typing import Any, Iterator, Optional, Tuple, Type, Union
 
 import numpy as np
 import torch
+from torch.utils.data._utils.pin_memory import pin_memory
 
 
 class BaseDataElement:
@@ -560,6 +561,16 @@ class BaseDataElement:
             k: v.to_dict() if isinstance(v, BaseDataElement) else v
             for k, v in self.all_items()
         }
+
+    def pin_memory(self) -> 'BaseDataElement':
+        """BaseDataElement supports pin_memory function to speed up
+        training."""
+        for k, v in self.items():
+            if not isinstance(v, BaseDataElement):
+                setattr(self, k, pin_memory(v))
+            else:
+                setattr(self, k, v.pin_memory())
+        return self
 
     def __repr__(self) -> str:
         """Represent the object."""
--- a/tests/test_structures/test_data_element.py
+++ b/tests/test_structures/test_data_element.py
@@ -458,3 +458,16 @@ class TestBaseDataElement(TestCase):
         metainfo, data = self.setup_data()
         instances = BaseDataElement(metainfo=metainfo, **data)
         self.assertDictEqual(instances.metainfo, metainfo)
+
+    def test_pin_memory(self):
+        metainfo, data = self.setup_data()
+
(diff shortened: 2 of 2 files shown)


mmengine/structures/base_data_element.py
```diff
@@ -5,6 +5,7 @@
 
 import numpy as np
 import torch
+from torch.utils.data._utils.pin_memory import pin_memory
```
RangiLyu, line 8: it is not safe to call a private method

mmengine/structures/base_data_element.py
```diff
@@ -561,6 +562,16 @@ def to_dict(self) -> dict:
             for k, v in self.all_items()
         }
 
+    def pin_memory(self) -> 'BaseDataElement':
+        """BaseDataElement supports pin_memory function to speed up
```
RangiLyu, line 566: need to indicate what is pin_memory in the docstring

mmengine/structures/base_data_element.py
```diff
@@ -561,6 +562,16 @@ def to_dict(self) -> dict:
...
+        training."""
+        for k, v in self.items():
+            if not isinstance(v, BaseDataElement):
```
RangiLyu, line 569: can simply replace with if hasattr(v, 'pin_memory') 
just follow the `cuda()` and `to()` method

mmengine/structures/base_data_element.py
```diff
@@ -561,6 +562,16 @@ def to_dict(self) -> dict:
             for k, v in self.all_items()
         }
 
+    def pin_memory(self) -> 'BaseDataElement':
```
RangiLyu, line 565: Need to provide `is_pinned` method

mmengine/structures/base_data_element.py
hhaAndroid: If you can't call it, we can only rewrite it?

mmengine/structures/base_data_element.py
hhaAndroid: There are many properties inside this object, so if one is not pinned, return false?

mmengine/structures/base_data_element.py
hhaAndroid: ok

mmengine/structures/base_data_element.py
HAOCHENYE: If we use the internal function `pin_memory`, seems this if-else is unnecessary. `pin_memory` will handle  various types of inputs.

mmengine/structures/base_data_element.py
```diff
@@ -561,6 +562,16 @@ def to_dict(self) -> dict:
...
+        """BaseDataElement supports pin_memory function to speed up
+        training."""
+        for k, v in self.items():
+            if not isinstance(v, BaseDataElement):
+                setattr(self, k, pin_memory(v))
```
HAOCHENYE, line 570: Suggest using `self.set_field`

PR #929, opened 2023-02-14
[Fix] Support configure torch.backends.cudnn.deterministic when CUDA>=10.2

HAOCHENYE:

## Motivation

Runner will raise an error if `deterministic` is set to False when CUDA >=10.2. According to the [official guide](https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility), set `CUBLAS_WORKSPACE_CONFIG=:16:8` in the environment variable could avoid the error (may limit overall performance).

```bash
RuntimeError: Deterministic behavior was enabled with either `torch.use_deterministic_algorithms(True)` or `at::Context::setDeterministicAlgorithms(true)`, but this operation is not deterministic because it uses CuBLAS and you have CUDA >= 10.2. To enable deterministic behavior in this case, you must set an environment variable before running your PyTorch application: CUBLAS_WORKSPACE_CONFIG=:4096:8 or CUBLAS_WORKSPACE_CONFIG=:16:8. For more information, go to https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility
```

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/runner/utils.py
+++ b/mmengine/runner/utils.py
@@ -1,5 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import logging
+import os
 import random
 from typing import List, Optional, Tuple
 
@@ -71,11 +72,14 @@ def set_random_seed(seed: Optional[int] = None,
     torch.cuda.manual_seed_all(seed)
     # os.environ['PYTHONHASHSEED'] = str(seed)
     if deterministic:
+        if digit_version(torch.version.cuda) >= digit_version('10.2'):
+            os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':16:8'
         if torch.backends.cudnn.benchmark:
             print_log(
                 'torch.backends.cudnn.benchmark is going to be set as '
                 '`False` to cause cuDNN to deterministically select an '
-                'algorithm',
+                'algorithm, which may limit overall performance, see more '
+                'informaton in https://docs.nvidia.com/cuda/cublas/index.html#results-reproducibility',  # noqa: E501
                 logger='current',
                 level=logging.WARNING)
         torch.backends.cudnn.deterministic = True


mmengine/runner/utils.py
```diff
@@ -71,11 +72,14 @@ def set_random_seed(seed: Optional[int] = None,
     torch.cuda.manual_seed_all(seed)
     # os.environ['PYTHONHASHSEED'] = str(seed)
     if deterministic:
+        if digit_version(torch.version.cuda) >= digit_version('10.2'):
+            os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':16:8'
```
C1rN09, line 76: Should add some references/explanations here

C1rN09 (approved):

mmengine/runner/utils.py
```diff
@@ -75,12 +76,21 @@ def set_random_seed(seed: Optional[int] = None,
...
-        if digit_version(TORCH_VERSION) >= digit_version('1.10.0'):
+        if digit_version(TORCH_VERSION) >= digit_version('1.8.0'):
+            # CUBLAS_WORKSPACE_CONFIG must be set for cudatoolkit version
+            # higher than 10.2 if deterministic is True.
+            if digit_version(torch.version.cuda) >= digit_version('10.2'):
+                # set a debug environment variable CUBLAS_WORKSPACE_CONFIG to
+                # :16:8 (may limit overall performance) or :4096:8 (will
+                # increase library footprint in GPU memory by approximately
+                # 24MiB).
+                os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':16:8'
```
zhouzaida, line 94: Why do we not set it as `:4096:8` which only increases 24MiB?

HAOCHENYE: Since the additionally occupied memory could case the OOM

zhouzaida: If users has set this environment, we should not update it.

zhouzaida: We also need to provide the benchmark of `:16:8` and `4096:8`.

HAOCHENYE:
waiting for https://github.com/pytorch/pytorch/issues/96724

PR #948, opened 2023-02-22
[Enhance] Make sure Tensors to broadcast is contiguous

XWHtorrentx:
This is linked to “To avoid tensor contiguity problems in distributed training #879”

### DIFF
--- a/mmengine/dist/dist.py
+++ b/mmengine/dist/dist.py
@@ -307,8 +307,8 @@ def broadcast(data: Tensor,
         input_device = get_data_device(data)
         backend_device = get_comm_device(group)
         data_on_device = cast_data_device(data, backend_device)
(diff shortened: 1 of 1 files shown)


mmengine/dist/dist.py
```diff
@@ -307,7 +307,8 @@ def broadcast(data: Tensor,
         input_device = get_data_device(data)
         backend_device = get_comm_device(group)
         data_on_device = cast_data_device(data, backend_device)
-
+        if isinstance(data_on_device,torch.Tensor):#this 'if' is to make lint happy
+            data_on_device=data_on_device.contiguous()
```
zhouzaida, line 311:
```suggestion
        # broadcast requires tensor is contiguous
        data_on_device = data_on_device.contiguous()  # type: ignore
```

PR #971, opened 2023-02-28
[Enhance] Add error trackback in Runner.call_hooks

Ezra-Yu:
## Motivation

Print error trackback in Runner.call_hooks if failure to help debug

## Modification

print trackback message before raising the TypeError.

## Use cases (Optional)

When debugging precise BN hook:

#### orginary trackback
image

#### add trackback
image

## Checklist

### DIFF
--- a/mmengine/runner/runner.py
+++ b/mmengine/runner/runner.py
@@ -6,9 +6,11 @@ import os.path as osp
 import pickle
 import platform
 import time
+import traceback
 import warnings
 from collections import OrderedDict
 from functools import partial
+from sys import stderr
 from typing import Callable, Dict, List, Optional, Sequence, Union
 
 import torch
@@ -1747,6 +1749,8 @@ class Runner:
                 try:
                     getattr(hook, fn_name)(self, **kwargs)
                 except TypeError as e:
+                    trace_msg = traceback.format_exc()
+                    print(f'\033[31m{trace_msg}\033[0m', file=stderr)
                     raise TypeError(f'{e} in {hook}') from None
 
     def register_hook(


HAOCHENYE:
How about just removing this annoying try-catch? The raised error mostly cannot tell where is the error triggered.

Ezra-Yu:
> How about just removing this annoying try-catch? The raised error mostly cannot tell where is the error triggered.

It seems that try-catch is useful. the unit tests fail.

HAOCHENYE:
We should preserve the check: `if hasattr(hook, func_name)`, otherwise users cannot set a custom point for the hook.

HAOCHENYE (approved):

zhouzaida (approved):

RangiLyu:
This try-catch is introduced by https://github.com/open-mmlab/mmengine/pull/247 to identify the hook name.

Ezra-Yu:
> This try-catch is introduced by #247 to identify the hook name.

If use `try-catch`, it will only give the info about which hook has an error, but no info about which section which line code caused the error. It makes users hard to debug. And you can look at track-back stack to loacte the error hook, just like:
image


Perhaps this commit option could also be used，https://github.com/open-mmlab/mmengine/pull/971/commits/98b7e6a9c8a6d6f258a411d772360a0d04458a7b

Ezra-Yu:
> where can I found the hook name 

The stack will return the file path and you can locate which hook it is by the filename. When there is an error, we don't know if it is a parameter error or a code error, so I don't think it is enough to just report the hook typeerror.
image

PR #979, opened 2023-03-06
[Fix] fix pytorch2.0 not support sys._getframe to remove private key

Harold-lkk:
Same for #690 
PyTorch2.0 does not support sys._getframe, so we remove the redundant inner attribute created by the property function by removing the `_` + property_name at the `keys()` function.

### DIFF
--- a/mmengine/structures/base_data_element.py
+++ b/mmengine/structures/base_data_element.py
@@ -1,6 +1,5 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import copy
-import sys
 from typing import Any, Iterator, Optional, Tuple, Type, Union
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


C1rN09 (approved):

mmengine/structures/base_data_element.py
```diff
@@ -309,7 +308,15 @@ def keys(self) -> list:
         Returns:
             list: Contains all keys in data_fields.
         """
-        return list(self._data_fields)
+        # We assume that the name of the attribute related to property is
+        # '_' + the name of the property. We use this rule to filter out
+        # private keys.
```
zhouzaida, line 313: Suggest adding a comment `# TODO: Use a more elegant way to solve this problem`.

zhouzaida, line 313:
```suggestion
        # private keys.
        # TODO: Use a more robust way to solve this problem
```

zhouzaida (approved):

Issue #517, opened 2022-09-07
`SyncBufferHook` does not work properly in `IterBasedTrainLoop` 


HAOCHENYE:
![image](https://user-images.githubusercontent.com/57566630/188780283-69bafcd6-9db4-44ea-9a1a-45198f72dce1.png)
Buffer should be synced before validation

PR #982, opened 2023-03-07
[Fix] Call SyncBufferHook before validation in IterBasedTrainLoop

Luo-Yihang:

## Motivation

This PR is tring to fix the issue mentioned in #517.

## Modification

Add a call to sync buffer before each validation.

## Checklist

### DIFF
--- a/mmengine/hooks/sync_buffer_hook.py
+++ b/mmengine/hooks/sync_buffer_hook.py
@@ -25,7 +25,7 @@ class SyncBuffersHook(Hook):
             if not self.called_in_train:
                 all_reduce_params(runner.model.buffers(), op='mean')
             self.called_in_train = False
-                
+
     def after_train_epoch(self, runner) -> None:
         """All-reduce model buffers at the end of each epoch.
 
--- a/mmengine/hooks/sync_buffer_hook.py
+++ b/mmengine/hooks/sync_buffer_hook.py
@@ -41,3 +41,14 @@ class SyncBuffersHook(Hook):
         if self.distributed:
             all_reduce_params(runner.model.buffers(), op='mean')
             self.called_in_train = True
+
+    def before_test_epoch(self, runner) -> None:
+        """All-reduce model buffers before each test epoch.
+
+        Args:
+            runner (Runner): The runner of the training process.
+        """
+        if self.distributed:
+            if not self.called_in_train:
+                all_reduce_params(runner.model.buffers(), op='mean')
+            self.called_in_train = False
--- a/mmengine/hooks/sync_buffer_hook.py
+++ b/mmengine/hooks/sync_buffer_hook.py
@@ -16,7 +16,9 @@ class SyncBuffersHook(Hook):
         self.called_in_train = False
 
     def before_val_epoch(self, runner) -> None:
-        """All-reduce model buffers before each validation epoch.
+        """All-reduce model buffers before each validation epoch. Sync the
+        buffer before each validation if it has not been synced at the end of
+        previous training epoch.
 
         Args:
             runner (Runner): The runner of the training process.
--- a/mmengine/hooks/sync_buffer_hook.py
+++ b/mmengine/hooks/sync_buffer_hook.py
@@ -13,6 +13,8 @@ class SyncBuffersHook(Hook):
 
     def __init__(self) -> None:
         self.distributed = is_distributed()
+        # A flag to mark whether synchronization has been done in
+        # after_train_epoch
         self.called_in_train = False
 
     def before_val_epoch(self, runner) -> None:
--- a/mmengine/hooks/sync_buffer_hook.py
+++ b/mmengine/hooks/sync_buffer_hook.py
@@ -13,7 +13,19 @@ class SyncBuffersHook(Hook):
 
     def __init__(self) -> None:
         self.distributed = is_distributed()
+        self.called_in_train = False
 
+    def before_val_epoch(self, runner) -> None:
+        """All-reduce model buffers before each validation epoch.
+
+        Args:
+            runner (Runner): The runner of the training process.
+        """
+        if self.distributed:
+            if not self.called_in_train:
+                all_reduce_params(runner.model.buffers(), op='mean')
+            self.called_in_train = False
+                
     def after_train_epoch(self, runner) -> None:
         """All-reduce model buffers at the end of each epoch.
 
@@ -22,3 +34,4 @@ class SyncBuffersHook(Hook):
         """
         if self.distributed:
             all_reduce_params(runner.model.buffers(), op='mean')
+            self.called_in_train = True
--- a/mmengine/hooks/sync_buffer_hook.py
+++ b/mmengine/hooks/sync_buffer_hook.py
@@ -41,14 +41,3 @@ class SyncBuffersHook(Hook):
         if self.distributed:
             all_reduce_params(runner.model.buffers(), op='mean')
             self.called_in_train = True
-
-    def before_test_epoch(self, runner) -> None:
-        """All-reduce model buffers before each test epoch.
-
-        Args:
-            runner (Runner): The runner of the training process.
-        """
-        if self.distributed:
-            if not self.called_in_train:
-                all_reduce_params(runner.model.buffers(), op='mean')
-            self.called_in_train = False
--- a/mmengine/hooks/sync_buffer_hook.py
+++ b/mmengine/hooks/sync_buffer_hook.py
@@ -18,9 +18,11 @@ class SyncBuffersHook(Hook):
         self.called_in_train = False
 
     def before_val_epoch(self, runner) -> None:
-        """All-reduce model buffers before each validation epoch. Sync the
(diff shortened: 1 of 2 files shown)


zhouzaida:
Hi @Luo-Yihang , thanks for your contribution.

mmengine/runner/loops.py
```diff
@@ -281,6 +281,7 @@ def run(self) -> None:
             if (self.runner.val_loop is not None
                     and self._iter >= self.val_begin
                     and self._iter % self.val_interval == 0):
+                self.runner.call_hook('after_train_epoch')
```
zhouzaida, line 284: This line will call all the hooks, which is not expected.
We can implement a `before_val_epoch` method in [SyncBuffersHook](https://github.com/open-mmlab/mmengine/blob/main/mmengine/hooks/sync_buffer_hook.py) which does the same action with `after_train_epoch`.
In addition, we can also add an attribute `called_in_train` to `SyncBuffersHook` which will be set as True in `after_train_epoch` after `synchronization` and set as False in `before_val_epoch`.

Luo-Yihang: @zhouzaida Thank you for such detailed explanation. May I ask the purpose of adding the attribute `called_in_train`, as it has not been used anywhere.

zhouzaida: Adding `called_in_train` is to avoid synchronizing the buffers twice in EpochBasedTrainLoop which will sync buffers in `after_train_epoch`.

https://github.com/open-mmlab/mmengine/blob/aeb5c454c56212eb1f94c5732de9be747d85ac87/mmengine/runner/loops.py#L107-L114

```python
@HOOKS.register_module()
class SyncBuffersHook(Hook):
    """Synchronize model buffers such as running_mean and running_var in BN at
    the end of each epoch."""

    priority = 'NORMAL'

    def __init__(self) -> None:
        self.distributed = is_distributed()
        self.called_in_train = False

    def after_train_epoch(self, runner) -> None:
        """All-reduce model buffers at the end of each epoch.
        Args:
            runner (Runner): The runner of the training process.
        """
        if self.distributed:
            all_reduce_params(runner.model.buffers(), op='mean')
            self.called_in_train = True
            

    def before_val_epoch(self, runner) -> None:
        """All-reduce model buffers at the end of each epoch.
        Args:
            runner (Runner): The runner of the training process.
        """
        if self.distributed and not self.called_in_train:
            all_reduce_params(runner.model.buffers(), op='mean')
            self.called_in_train = False
```

Luo-Yihang: @zhouzaida Thanks! I have made changes accordingly. 
There is only one thing left to discuss. For the implementation of `before_val_epoch`, I made `self.called_in_train = False` executed whatever the `called_in_train` is. Please correct me if I am mistaken.

mmengine/hooks/sync_buffer_hook.py
```diff
@@ -13,6 +13,18 @@ class SyncBuffersHook(Hook):
...
     def __init__(self) -> None:
         self.distributed = is_distributed()
+        self.called_in_train = False
+
+    def before_val_epoch(self, runner) -> None:
+        """All-reduce model buffers before each validation epoch.
+
+        Args:
+            runner (Runner): The runner of the training process.
+        """
```
zhouzaida, line 23: Suggest adding a short description to explain why we need to implement `before_val_epoch` method.

Luo-Yihang: I have added such description, please check on that. Thanks!

mmengine/hooks/sync_buffer_hook.py
```diff
@@ -13,6 +13,20 @@ class SyncBuffersHook(Hook):
...
+        self.called_in_train = False
+
+    def before_val_epoch(self, runner) -> None:
+        """All-reduce model buffers before each validation epoch. Sync the
+        buffer before each validation if it has not been synced at the end of
+        previous training epoch.
+
+        Args:
+            runner (Runner): The runner of the training process.
+        """
```
zhouzaida, line 25:
```suggestion
        """All-reduce model buffers before each validation epoch.
        
        Synchronize the buffers before each validation if they have not been
        synchronized at the end of the previous training epoch. This method
        will be called when using IterBasedTrainLoop.

        Args:
            runner (Runner): The runner of the training process.
        """
```

mmengine/hooks/sync_buffer_hook.py
```diff
@@ -13,6 +13,20 @@ class SyncBuffersHook(Hook):
 
     def __init__(self) -> None:
         self.distributed = is_distributed()
+        self.called_in_train = False
```
zhouzaida, line 16:
```suggestion
        # A flag to mark whether synchronization has been done in
        # after_train_epoch
        self.called_in_train = False
```

mmengine/hooks/sync_buffer_hook.py
```diff
@@ -13,6 +13,24 @@ class SyncBuffersHook(Hook):
 
     def __init__(self) -> None:
         self.distributed = is_distributed()
+        # A flag to mark whether synchronization has been done in
+        # after_train_epoch
+        self.called_in_train = False
```
HAOCHENYE, line 18: We could set `self.called_in_train=True` by default. Then the unnecessary buffer synchronization will not happen when we call `runner.val()` separately.

Luo-Yihang: Looks like such setting will make buffer sync before validation in IterBasedTrainLoop not work. Correct me if I am wrong.

HAOCHENYE: Ahh, you're right, just ignore this comment!

zhouzaida:
The final comment is that maybe it is necessary also to implement a `before_test_epoch`.

Luo-Yihang:
> The final comment is that maybe it is necessary also to implement a `before_test_epoch`.

Hi, please check the new commit for this.

Luo-Yihang:
Hi, is there any updates on this PR?

mmengine/hooks/sync_buffer_hook.py
```diff
@@ -13,6 +13,24 @@ class SyncBuffersHook(Hook):
...
+        synchronized at the end of the previous training epoch. This method
+        will be called when using IterBasedTrainLoop.
+
+        Args:
+            runner (Runner): The runner of the training process.
+        """
+        if self.distributed:
+            if not self.called_in_train:
+                all_reduce_params(runner.model.buffers(), op='mean')
+            self.called_in_train = False
```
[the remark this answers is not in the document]
Luo-Yihang: But `all_reduce_params` will only be called when `self.called_in_train` is False. I think we need to always set `self.called_in_train` to False here. Otherwise `self.called_in_train` will always be True and never get updated.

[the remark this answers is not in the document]
zhouzaida: `self.called_in_train` is set as False in `__init__` so `all_reduce_params` will be called in `before_val_epoch` when using `IterBasedTrainLoop`. After `all_reduce_params` is called, the `self.called_in_train` will always be False.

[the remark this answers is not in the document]
Luo-Yihang: Yes, I agree. I believe both implementation will work correctly in `IterBasedTrainLoop`(and all other cases we have now). But could you please briefly explain why its necessary to only set `self.called_in_train` as False when `all_reduce_params`  is called. Is this for code clarity or something?  As I think always setting self.called_in_train to False may seem like a safer approach.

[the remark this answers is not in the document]
zhouzaida: It may look clear but your implementation is also correct so there is no need to update. Thanks for your contribution.

HAOCHENYE (approved):

mmengine/hooks/sync_buffer_hook.py
```diff
@@ -22,3 +40,15 @@ def after_train_epoch(self, runner) -> None:
         """
         if self.distributed:
             all_reduce_params(runner.model.buffers(), op='mean')
+            self.called_in_train = True
+
+    def before_test_epoch(self, runner) -> None:
```
zhouzaida, line 45: Sorry for my missing details. This method is not necessary to be implemented because `after_train_epoch` would sync the buffers for both `IterBasedTrainLoop` and `EpochBasedTrainLoop`.

Luo-Yihang: Do you mean that `before_test_epoch` is not neccessary?

zhouzaida: Yes

zhouzaida (approved):
@Luo-Yihang Thanks for your contribution. LGTM.

Luo-Yihang:
Hi, is there any updates on this PR?

zhouzaida:
> Hi, is there any updates on this PR?

Hi @Luo-Yihang , thanks for your contribution. This PR has been merged.

PR #992, opened 2023-03-10
[Fix] Fix repeated warning

HAOCHENYE:

## Motivation

`MMLogger.warning` will not filter the repeated warning, therefore the training message could be fully filled with the signs.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/logging/logger.py
+++ b/mmengine/logging/logger.py
@@ -18,9 +18,9 @@ class UniqueWarningFilter(logging.Filter):
         name (str): name of the filter.
     """
 
-    def __init__(self, name='mmengine'):
+    def __init__(self, name: str = 'mmengine'):
         super().__init__(name)
-        self.record_dict = set()
+        self.record_dict: set = set()
 
     def filter(self, record: LogRecord) -> bool:
         """Filter the repeated warning message.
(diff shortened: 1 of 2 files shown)


mmengine/logging/logger.py
```diff
@@ -11,6 +11,36 @@
 from mmengine.utils.manager import _accquire_lock, _release_lock
 
 
+class UniqueWarningFilter(logging.Filter):
```
zhouzaida, line 14:
```suggestion
class FilterDuplicateWarning(logging.Filter):
```

mmengine/logging/logger.py
```diff
@@ -11,6 +11,36 @@
...
+        Returns:
+            bool: Whether to output the log record.
+        """
+        if record.levelno != logging.WARNING:
+            return True
+
+        msg = record.msg
+        if msg not in self.seen:
+            self.seen.add(msg)
+            return True
```
zhouzaida, line 40:
```suggestion
        if record.msg not in self.seen:
            self.seen.add(record.msg)
            return True
```

zhouzaida (approved):

PR #997, opened 2023-03-14
[Enhancement] Unify the use of print_log and logger

LEFTeyex:

## Motivation

## Modification

There are 2 cases:
1. Use print_log, when there is no logger/self.logger.
2. Use logger/self.logger, when print_log and logger/self.logger are present at the same time.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

3. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
4. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
5. The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/mmengine/model/base_module.py
+++ b/mmengine/model/base_module.py
@@ -95,15 +95,12 @@ class BaseModule(nn.Module, metaclass=ABCMeta):
             for sub_module in self.modules():
                 sub_module._params_init_info = self._params_init_info
 
-        logger = MMLogger.get_current_instance()
-        logger_name = logger.instance_name
-
         module_name = self.__class__.__name__
         if not self._is_init:
             if self.init_cfg:
                 print_log(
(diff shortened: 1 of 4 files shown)


mmengine/model/base_module.py
```diff
@@ -176,10 +167,9 @@ def _dump_init_info(self):
                 with_file_handler = True
         if not with_file_handler:
             for name, param in self.named_parameters():
-                print_log(
+                logger.info(
```
zhouzaida, line 170: Why do we replace `print_log` with `logger.info`?

HAOCHENYE: Maybe here we must get the logger to use the handler to write messages to files. Since we've got logger, there is not need to use `print_log` anymore.

HAOCHENYE (approved):

PR #1011, opened 2023-03-24
[Enhancement] Support BoolTensor and LongTensor on Ascend NPU

Ginray:
## Motivation

When we test the ssd model on the NPU, an assert error occurs during the eval phase, because npu.BoolTensor and npu.LongTensor are not supported in instane_data.py.

## Modification

Modify the instane_data.py file to support NPU BoolTensor and LongTensor.

## BC-breaking (Optional)

Not affected.

## Use cases (Optional)

Not affected.

### DIFF
--- a/mmengine/structures/instance_data.py
+++ b/mmengine/structures/instance_data.py
@@ -1,16 +1,26 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import itertools
 from collections.abc import Sized
-from typing import List, Union
+from typing import Any, List, Union
 
 import numpy as np
 import torch
 
+from mmengine.device import get_device
 from .base_data_element import BaseDataElement
 
-IndexType = Union[str, slice, int, list, torch.LongTensor,
-                  torch.cuda.LongTensor, torch.BoolTensor,
-                  torch.cuda.BoolTensor, np.ndarray]
+BoolTypeTensor: Union[Any]
+LongTypeTensor: Union[Any]
+
+if get_device() == 'npu':
+    BoolTypeTensor = Union[torch.BoolTensor, torch.npu.BoolTensor]
+    LongTypeTensor = Union[torch.LongTensor, torch.npu.LongTensor]
+else:
+    BoolTypeTensor = Union[torch.BoolTensor, torch.cuda.BoolTensor]
+    LongTypeTensor = Union[torch.LongTensor, torch.cuda.LongTensor]
+
+IndexType: Union[Any] = Union[str, slice, int, list, LongTypeTensor,
+                              BoolTypeTensor, np.ndarray]
 
 
 # Modified from
@@ -156,6 +166,7 @@ class InstanceData(BaseDataElement):
         Returns:
             :obj:`InstanceData`: Corresponding values.
         """
+        assert isinstance(item, IndexType.__args__)
         if isinstance(item, list):
             item = np.array(item)
         if isinstance(item, np.ndarray):
@@ -165,9 +176,6 @@ class InstanceData(BaseDataElement):
             # More details in https://github.com/numpy/numpy/issues/9464
             item = item.astype(np.int64) if item.dtype == np.int32 else item
             item = torch.from_numpy(item)
-        assert isinstance(
-            item, (str, slice, int, torch.LongTensor, torch.cuda.LongTensor,
-                   torch.BoolTensor, torch.cuda.BoolTensor))
 
         if isinstance(item, str):
             return getattr(self, item)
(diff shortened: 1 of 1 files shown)


mmengine/structures/instance_data.py
```diff
@@ -165,9 +172,8 @@ def __getitem__(self, item: IndexType) -> 'InstanceData':
             # More details in https://github.com/numpy/numpy/issues/9464
             item = item.astype(np.int64) if item.dtype == np.int32 else item
             item = torch.from_numpy(item)
-        assert isinstance(
-            item, (str, slice, int, torch.LongTensor, torch.cuda.LongTensor,
-                   torch.BoolTensor, torch.cuda.BoolTensor))
+        assert isinstance(item, (str, slice, int, LongTypeTensor.__args__,
+                                 BoolTypeTensor.__args__))
```
HAOCHENYE, line 176:
```suggestion
        assert isinstance(item, (str, slice, int, typing.get_args(LongTypeTensor), typing.get_args(BoolTypeTensor))
```

mmengine/structures/instance_data.py
```diff
@@ -183,7 +189,7 @@ def __getitem__(self, item: IndexType) -> 'InstanceData':
         if isinstance(item, torch.Tensor):
             assert item.dim() == 1, 'Only support to get the' \
                                     ' values along the first dimension.'
-            if isinstance(item, (torch.BoolTensor, torch.cuda.BoolTensor)):
+            if isinstance(item, BoolTypeTensor.__args__):
```
HAOCHENYE, line 192: same as above

mmengine/structures/instance_data.py
Ginray: Thanks for your suggestion, I found out that "typing.get_args" method only enabled when python>= 3.8. Do we need compatibility with older versions ?

zhouzaida: We also need to support Python3.7.

mmengine/structures/instance_data.py
```diff
@@ -1,16 +1,23 @@
...
 
-IndexType = Union[str, slice, int, list, torch.LongTensor,
-                  torch.cuda.LongTensor, torch.BoolTensor,
-                  torch.cuda.BoolTensor, np.ndarray]
+BoolTypeTensor: Union[Any] = Union[torch.BoolTensor, torch.cuda.BoolTensor,
+                                   torch.npu.BoolTensor] if \
+    is_npu_available() else Union[torch.BoolTensor, torch.cuda.BoolTensor]
+LongTypeTensor: Union[Any] = Union[torch.LongTensor, torch.cuda.LongTensor,
+                                   torch.npu.LongTensor] if \
+    is_npu_available() else Union[torch.LongTensor, torch.cuda.LongTensor]
```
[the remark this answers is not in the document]
Ginray: Fixed according to review comments.

HAOCHENYE (approved):

mmengine/structures/instance_data.py
```diff
@@ -1,16 +1,28 @@
...
-                  torch.cuda.LongTensor, torch.BoolTensor,
-                  torch.cuda.BoolTensor, np.ndarray]
+BoolTypeTensor: Union[Any]
+LongTypeTensor: Union[Any]
+
+if get_device() == 'npu':
+    BoolTypeTensor = Union[torch.BoolTensor, torch.cuda.BoolTensor,
+                           torch.npu.BoolTensor]
+    LongTypeTensor = Union[torch.LongTensor, torch.cuda.LongTensor,
+                           torch.npu.LongTensor]
```
zhouzaida, line 19: Is it necessary to keep `torch.cuda.BoolTensor` when the current_device is 'npu'?

mmengine/structures/instance_data.py
```diff
@@ -165,9 +177,8 @@ def __getitem__(self, item: IndexType) -> 'InstanceData':
             # More details in https://github.com/numpy/numpy/issues/9464
             item = item.astype(np.int64) if item.dtype == np.int32 else item
             item = torch.from_numpy(item)
-        assert isinstance(
-            item, (str, slice, int, torch.LongTensor, torch.cuda.LongTensor,
-                   torch.BoolTensor, torch.cuda.BoolTensor))
+        assert isinstance(item, (str, slice, int, LongTypeTensor.__args__,
+                                 BoolTypeTensor.__args__))
```
zhouzaida, line 181:
```suggestion
        assert isinstance(item, IndexType.__args__)
```

zhouzaida: And move this line to Line 171.

```python
assert isinstance(item, IndexType.__args__)
if isinstance(item, list):
    item = np.array(item)
if isinstance(item, np.ndarray):
    # The default int type of numpy is platform dependent, int32 for
    # windows and int64 for linux. `torch.Tensor` requires the index
    # should be int64, therefore we simply convert it to int64 here.
    # More details in https://github.com/numpy/numpy/issues/9464
    item = item.astype(np.int64) if item.dtype == np.int32 else item
    item = torch.from_numpy(item)
```

mmengine/structures/instance_data.py
Ginray: Not necessary, already fixed.

mmengine/structures/instance_data.py
Ginray: Thanks for the suggestion, already fixed.

zhouzaida (approved):

PR #1036, opened 2023-04-01
[Enhance] Distinguish training/testing saving directory

shufanwu:

## Motivation

Distinguish training/testing saving directory.

## Modification

add suffix for log_dir in Runner after call train() val() test() functions.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/runner/runner.py
+++ b/mmengine/runner/runner.py
@@ -270,6 +270,8 @@ class Runner:
         experiment_name: Optional[str] = None,
         cfg: Optional[ConfigType] = None,
     ):
+        # Record call sequence of task processes
+        self._task_list: List[str] = []
         self._work_dir = osp.abspath(work_dir)
         mmengine.mkdir_or_exist(self._work_dir)
 
@@ -1712,6 +1714,7 @@ class Runner:
 
         model = self.train_loop.run()  # type: ignore
         self.call_hook('after_run')
+        self._task_list.append('train')
         return model
 
     def val(self) -> dict:
@@ -1735,6 +1738,7 @@ class Runner:
 
         metrics = self.val_loop.run()  # type: ignore
         self.call_hook('after_run')
+        self._task_list.append('val')
         return metrics
 
     def test(self) -> dict:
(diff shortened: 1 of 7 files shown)


mmengine/runner/runner.py
```diff
@@ -1643,6 +1644,21 @@ def load_or_resume(self) -> None:
...
+            task_type: task type the runner lunched (train val test)
+
+        Returns:
+
+        """
+        assert task_type in ('train', 'val', 'test'), (
+            'the value of task type must be one of ["train", "val", "test"]')
+        previous_log_dir = self._log_dir
+        self._log_dir = '_'.join([self._log_dir, task_type])
+        shutil.move(previous_log_dir, self._log_dir)
```
HAOCHENYE, line 1660: Hi, will the `FileHandler` in the logger still save logs to the old log file?

shufanwu: Yes, I can fix this if it should modify the log file in FileHandles, and I wonder whether the save_dir in visualizer should modify in the same time.

HAOCHENYE: Yep, all opened files should be closed and be redirected to the new directory.

shufanwu: If we close the files opened by `TensorboardVisBackend`, when we want to use it once again, we must to create a new instance of `TensorboardVisBackend` and it may create two events.out file in the log dir. Is this what we expect?

HAOCHENYE: Emm, resuming a visualization seems weird and dirty more or less. A more reasonable workaround is that we should lazy initialize the visualizer in `test`, `train`, `val` rather than `__init__` , but I'm not sure will it has a side effect.

Besides, if I first run `runner.train()`, and then run `runer.test()`, does it mean we need to create a new visualizer and logger during `test`? What is the lifespan of the logger and visualization? Maybe we need to rethink it carefully. What do you think of it?

HAOCHENYE: I propose the following approach: We maintain the current logic of `runner.train`, keeping the output log file names unchanged. However, we add a condition in `runner.test`. If the train_loop has already been initialized (indicating that testing is being conducted immediately after training), we will not change the log path. On the other hand, if the train_loop has not been initialized (meaning the test is being run directly), we can simply rename the folder at the end of the testing process. This way, there is no need to modify the output paths for FileHandler and Visualizer. Does this approach sound reasonable to you?

PR #1039, opened 2023-04-02
[Draft] Add fsdp strategy

HAOCHENYE:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/strategy/fsdp_strategy.py
+++ b/mmengine/strategy/fsdp_strategy.py
@@ -68,6 +68,7 @@ class FSDPStrategy(DDPStrategy):
             'Model should have been built before wrap')
 
         self.fsdp_kwargs.setdefault('type', 'MMFullyShardedDataParallel')
+        self.fsdp_kwargs.setdefault('module', self.model)
         model_wrapper_cfg: dict = self.cfg.get('model_wrapper_cfg', dict())
         if self.fsdp_kwargs:
             assert not model_wrapper_cfg, (
--- a/mmengine/strategy/fsdp_strategy.py
+++ b/mmengine/strategy/fsdp_strategy.py
@@ -1,0 +1,144 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import logging
+import os
+import os.path as osp
+from copy import deepcopy
+from typing import Dict, Optional, Union
+
+from torch.distributed.fsdp import FullStateDictConfig
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+from torch.distributed.fsdp import (LocalStateDictConfig,
+                                    ShardedStateDictConfig, StateDictType)
+from torch.distributed.fsdp.fully_sharded_data_parallel import (
+    FullOptimStateDictConfig, LocalOptimStateDictConfig,
+    ShardedOptimStateDictConfig)
+
+from mmengine.dist import get_rank, is_main_process
+from mmengine.model import is_model_wrapper
+from mmengine.registry import MODEL_WRAPPERS, STRATEGIES, Registry
+from mmengine.utils import mkdir_or_exist
+from .ddp_strategy import DDPStrategy
+
+FSDP_CONFIGS = Registry('fsdp configs')
+FSDP_CONFIGS.register_module(module=FullOptimStateDictConfig)
+FSDP_CONFIGS.register_module(module=ShardedOptimStateDictConfig)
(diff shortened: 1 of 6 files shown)


mmengine/strategy/fsdp_strategy.py
```diff
@@ -0,0 +1,144 @@
...
+        wrapper_type = wrapper_args.pop('type', 'MMFullyShardedDataParallel')
+        wrapper_cls = MODEL_WRAPPERS.get(wrapper_type)
+        self.model = wrapper_cls(
+            module=self.model,
+            **wrapper_args)
+        self.model.set_state_dict_type(
+            self.model,
+            self.state_dict_type,
+            self.state_dict_config,
+            self.optim_state_dict_config)
```
C1rN09, line 78: This API is available only when PyTorch >= 2.0.0

mmengine/strategy/fsdp_strategy.py
```diff
@@ -0,0 +1,144 @@
...
+from mmengine.utils import mkdir_or_exist
+from .ddp_strategy import DDPStrategy
+
+FSDP_CONFIGS = Registry('fsdp configs')
+FSDP_CONFIGS.register_module(module=FullOptimStateDictConfig)
+FSDP_CONFIGS.register_module(module=ShardedOptimStateDictConfig)
+FSDP_CONFIGS.register_module(module=LocalOptimStateDictConfig)
+FSDP_CONFIGS.register_module(module=FullStateDictConfig)
+FSDP_CONFIGS.register_module(module=ShardedStateDictConfig)
+FSDP_CONFIGS.register_module(module=LocalStateDictConfig)
```
C1rN09, line 28: I believe some of these are only available after PyTorch 2.0

mmengine/strategy/fsdp_strategy.py
```diff
@@ -0,0 +1,144 @@
...
+            **fsdp_kwargs):
+        self.state_dict_type = StateDictType[state_dict_type]
+        self.state_dict_config = FSDP_CONFIGS.build(state_dict_config)
+        self.optim_state_dict_config = FSDP_CONFIGS.build(optim_state_dict_config)
+        super(DDPStrategy, self).__init__(
+            logger=logger,
+            amp=amp,
+            accumulative_counts=accumulative_counts,
+            clip_grad=clip_grad,
+            **fsdp_kwargs)
```
C1rN09, line 56: Should `DDPStrategy` receive `**fsdp_kwargs`?

mmengine/strategy/native_strategy.py
```diff
@@ -170,39 +169,47 @@ def save_checkpoint(self,
...
-        from mmengine.runner.checkpoint import (_load_checkpoint,
-                                                _load_checkpoint_to_model)
-        if name is None:
+        if filepath is None:
             # should auto detect latest checkpoint
-            name = self._latest_checkpoint_name(load_dir)
-        if name is None:
+            filepath = self._latest_checkpoint_name(load_dir)
+        if filepath is None:
             return None
```
C1rN09, line 199: I have to clarify why this API seems kind of `weired`. It is designed to be compatible with [DeepSpeed](https://github.com/microsoft/DeepSpeed/blob/f1e4fb0b878e48d6fe33dbb2e0969d791d698e86/deepspeed/runtime/engine.py#LL3084C43-L3084C43) ones

```python
def save_checkpoint(self, save_dir, tag=None, client_state={}, save_latest=True):
    ....
```

1. It doesn't split `get_state_dict` and `save_checkpoint_to_file` into 2 stages, so you cannot get a state_dict first and then save it with your own method.
2. It saves the `tag` into a different file (`last_ckpt`, if I remembered well) than ours (`latest_checkpoint`, if I remembered well), and the `tag` is a relative path rather than a full path (that's what we did)
3. `tag` actually indicates a directory rather than a file.

For example, if you set `save_dir='aaa'` and `tag='bbb'` and `save_latest=True`, you will get something like:

```text
aaa
|- bbb
  |- model_rank_00.pth.
  |- model_rank_01.pth
  |- optim_rank_00.pth
  |- optim_rank_01.pth
|- last_ckpt  # a text file, where a single line of string 'bbb' is stored
```

And `load_checkpoint` is similar.

C1rN09 (commented):
Generally good. But FSDP's API is changing rapidly, so we should consider whether we only support FSDP from PyTorch 2.0.0 or an older version.

Issue #935, opened 2023-02-17
[Bug] No training log when the num of iterations smaller than default interval


Tau-J:
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmengine/issues) and [Discussions](https://github.com/open-mmlab/mmengine/discussions) but cannot get the expected help.
- [X] The bug has not been fixed in the latest version(https://github.com/open-mmlab/mmengine).

### Environment

```
mmengine == 0.5.0
```

### Reproduces the problem - code sample

-

### Reproduces the problem - command or script

-

### Reproduces the problem - error message

-

### Additional information

When the number of iterations per epoch is smaller than the log interval, there's no training info be printed. Since important information like `eta` is in training logs, I suggest printing at least one line of training logs per epoch.

Here's an example of current training logs:  
```
2023/01/31 13:44:41 - mmengine - INFO - Saving checkpoint at 200 epochs
2023/01/31 13:44:45 - mmengine - INFO - Evaluating MpiiPCKAccuracy (normalized by ``"head_size"``)...
2023/01/31 13:44:45 - mmengine - INFO - Epoch(val) [200][12/12]  Head PCK: 96.958305  Shoulder PCK: 96.375085  Elbow PCK: 90.350029  Wrist PCK: 85.442129  Hip PCK: 89.800562  Knee PCK: 86.871427  Ankle PCK: 82.686027  PCK: 90.257132  PCK@0.1: 33.794920
2023/01/31 13:45:05 - mmengine - INFO - Exp name: rtmpose-m_finetune-aic-coco_8xb64-210e_mpii_20230131_123528
2023/01/31 13:45:35 - mmengine - INFO - Exp name: rtmpose-m_finetune-aic-coco_8xb64-210e_mpii_20230131_123528
2023/01/31 13:45:56 - mmengine - INFO - Exp name: rtmpose-m_finetune-aic-coco_8xb64-210e_mpii_20230131_123528
2023/01/31 13:46:17 - mmengine - INFO - Exp name: rtmpose-m_finetune-aic-coco_8xb64-210e_mpii_20230131_123528
2023/01/31 13:46:30 - mmengine - INFO - Exp name: rtmpose-m_finetune-aic-coco_8xb64-210e_mpii_20230131_123528
2023/01/31 13:46:37 - mmengine - INFO - Exp name: rtmpose-m_finetune-aic-coco_8xb64-210e_mpii_20230131_123528
[...]

PR #1046, opened 2023-04-03
[Fix] No training log when the num of iterations is smaller than the default interval

shufanwu:

## Motivation

No training log when the num of iterations is smaller than the default interval. [https://github.com/open-mmlab/mmengine/issues/935](Issue)

## Modification

Add judgement in  function ```after_train_iter``` of class ```LoggerHook```.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/hooks/logger_hook.py
+++ b/mmengine/hooks/logger_hook.py
@@ -167,7 +167,8 @@ class LoggerHook(Hook):
             tag, log_str = runner.log_processor.get_log_after_iter(
                 runner, batch_idx, 'train')
         elif (self.end_of_epoch(runner.train_dataloader, batch_idx)
-              and not self.ignore_last):
+              and (not self.ignore_last
+                   or len(runner.train_dataloader) < self.interval)):
             # `runner.max_iters` may not be divisible by `self.interval`. if
             # `self.ignore_last==True`, the log of remaining iterations will
             # be recorded (Epoch [4][1000/1007], the logs of 998-1007
--- a/mmengine/hooks/logger_hook.py
+++ b/mmengine/hooks/logger_hook.py
@@ -168,7 +168,7 @@ class LoggerHook(Hook):
                 runner, batch_idx, 'train')
         elif (self.end_of_epoch(runner.train_dataloader, batch_idx)
               and (not self.ignore_last
-                   or len(runner.train_dataloader) < self.interval)):
+                   or len(runner.train_dataloader) <= self.interval)):
             # `runner.max_iters` may not be divisible by `self.interval`. if
             # `self.ignore_last==True`, the log of remaining iterations will
             # be recorded (Epoch [4][1000/1007], the logs of 998-1007
--- a/tests/test_runner/test_runner.py
+++ b/tests/test_runner/test_runner.py
@@ -1727,12 +1727,16 @@ class TestRunner(TestCase):
         with self.assertRaisesRegex(AssertionError, 'If you want to validate'):
             runner.train()
 
-        # 13 Test the logs will be output when the length of
-        # the train dataloader is smaller than the log interval
+        # 13 Test the logs will be printed when the length of
+        # train_dataloader is smaller than the interval set in LoggerHook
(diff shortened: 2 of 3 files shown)


HAOCHENYE (commented):
Hi, thanks for your contribution. The modification makes sense to me, but should we also update the `after_val_iter` and `after_test_iter`?

shufanwu:
> Hi, thanks for your contribution. The modification makes sense to me, but should we also update the `after_val_iter` and `after_test_iter`?
I think we don't need to  update these because in val or test task, the loggerhook `after_val_epoch` and  `after_test_epoch` will be called after each epoch.

HAOCHENYE:
> > Hi, thanks for your contribution. The modification makes sense to me, but should we also update the `after_val_iter` and `after_test_iter`?
> 
> I think we don't need to update these because in val or test task, the loggerhook `after_val_epoch` and `after_test_epoch` will be called after each epoch.

It makes sense to me :+1:

mmengine/hooks/logger_hook.py
```diff
@@ -167,7 +167,8 @@ def after_train_iter(self,
             tag, log_str = runner.log_processor.get_log_after_iter(
                 runner, batch_idx, 'train')
         elif (self.end_of_epoch(runner.train_dataloader, batch_idx)
-              and not self.ignore_last):
+              and (not self.ignore_last
+                   or len(runner.train_dataloader) < self.interval)):
```
HAOCHENYE, line 171:
```suggestion
                   or len(runner.train_dataloader) <= self.interval)):
```

HAOCHENYE:
Hi! Could you add a unit test for `LoggerHook` to validate that the logs will be output when the length of the train dataloader is smaller than the log interval?

HAOCHENYE:
Hi, it could be more proper to add unit test [here](https://github.com/open-mmlab/mmengine/blob/6b366f236c93d2735b336ee41dc521db1dd370e0/tests/test_hooks/test_logger_hook.py#L64), and we should test runner.logger.info will be called when the length of dataloader is smaller than the log interval.

tests/test_runner/test_runner.py
```diff
@@ -1727,6 +1727,13 @@ def train_step(self, *args, **kwargs):
         with self.assertRaisesRegex(AssertionError, 'If you want to validate'):
             runner.train()
 
+        # 13 Test the logs will be output when the length of
+        # the train dataloader is smaller than the log interval
+        cfg = copy.deepcopy(self.epoch_based_cfg)
+        cfg.experiment_name = 'test_train13'
+        runner = Runner.from_cfg(cfg)
+        runner.train()
```
zhouzaida, line 1735:
```suggestion
        # 13 Test the logs will be printed when the length of
        # train_dataloader is smaller than the interval set in LoggerHook
        cfg = copy.deepcopy(self.epoch_based_cfg)
        cfg.experiment_name = 'test_train13'
        cfg.default_hooks = dict(logger=dict(type='LoggerHook', interval=5))
        runner = Runner.from_cfg(cfg)
        runner.train()
        with open(runner.logger._log_file, 'r') as f:
            log = f.read()
        self.assertIn('Epoch(train) [1][4/4]', log)
```

zhouzaida (approved):

tests/test_runner/test_runner.py
```diff
@@ -1727,6 +1727,17 @@ def train_step(self, *args, **kwargs):
...
             runner.train()
 
+        # 13 Test the logs will be printed when the length of
+        # train_dataloader is smaller than the interval set in LoggerHook
+        cfg = copy.deepcopy(self.epoch_based_cfg)
+        cfg.experiment_name = 'test_train13'
+        cfg.default_hooks = dict(logger=dict(type='LoggerHook', interval=5))
+        runner = Runner.from_cfg(cfg)
+        runner.train()
+        with open(runner.logger._log_file, 'r') as f:
```
zhouzaida, line 1737:
```suggestion
        with open(runner.logger._log_file) as f:
```

zhouzaida (approved):

PR #595, opened 2022-10-11
[Feature] support registering partial functions and more

C1rN09:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

**This PR is converted to DRAFT because I think this may break the design of `Registry` and may mess up some concept. I will re-examine the necessity and rationality of this change.**

Support registering partial functions in MMEngine, solve #594 

This feature might be required by FSDP for `auto_wrap_policies`

## Modification

modify `Registry` to accept all `Callable`, including classes, functions, partial functions and more

## BC-breaking (Optional)

No

## Use cases (Optional)

```python
import functools
from mmengine.registry import Registry
FUNCS = Registry('funcs')
@FUNCS.register_module('resnet')
def build_resnet(layers=50, activation='relu'):
    pass
build_resnet18 = functools.partial(build_resnet, layers=18)
FUNCS.register_module('resnet18', False, build_resnet18)
model = FUNCS.build(dict(type='resnet18', activation='silu'))
```

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #1047, opened 2023-04-04
[Improve] Support registering method and partial function.

mzr1996:
## Motivation

The original registry doesn't accept method and partial functions.

## Modification

Use `callable` instead of `inspect.isfunction` as the condition of the registry.

## BC-breaking

No

## Use cases

For example, we want to use the factory method instead of the initialization method to build a class

```python
from transformers import AutoModelForImageClassification
from mmengine.registry import MODELS

MODELS.register_module('hf', module=AutoModelForImageClassification.from_pretrained)

resnet = MODELS.build(dict(type='hf', pretrained_model_name_or_path='microsoft/resnet-50'))
print(type(resnet))
# <class 'transformers.models.resnet.modeling_resnet.ResNetForImageClassification'>
```

## Checklist

### DIFF
--- a/mmengine/registry/registry.py
+++ b/mmengine/registry/registry.py
@@ -572,7 +572,7 @@ class Registry:
             force (bool): Whether to override an existing class with the same
                 name. Defaults to False.
         """
-        if not inspect.isclass(module) and not inspect.isfunction(module):
+        if not inspect.isclass(module) and not callable(module):
             raise TypeError('module must be a class or a function, '
                             f'but got {type(module)}')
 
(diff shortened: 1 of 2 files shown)


mmengine/registry/registry.py
```diff
@@ -572,7 +572,7 @@ def _register_module(self,
             force (bool): Whether to override an existing class with the same
                 name. Defaults to False.
         """
-        if not inspect.isclass(module) and not inspect.isfunction(module):
+        if not inspect.isclass(module) and not callable(module):
```
zhouzaida, line 575: Please update the message in TypeError. `a function` -> `a callable object`

zhouzaida:
closed by #595

Issue #1055, opened 2023-04-08
[Bug] The required arguments of `get_model_complexity_info()`


sjiang95:
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmengine/issues) and [Discussions](https://github.com/open-mmlab/mmengine/discussions) but cannot get the expected help.
- [X] The bug has not been fixed in the latest version(https://github.com/open-mmlab/mmengine).

### Environment

[...]

PR #1056, opened 2023-04-08
[Enhance] Make the parameters of get_model_complexity_info() friendly

sjiang95:
## Motivation

Optimize the flexibility of function `get_model_complexity_info()`. #1055

## Modification

1. Set default value of `input_shape` to `None`.
2. `assert` does not allow both `input_shape` and `inputs` are `None`.
3. If both `input_shape` and `inputs` are provided, print a warning.

## BC-breaking (Optional)

It the downstream repo calls the function `get_model_complexity_info()` like [this line](https://github.com/open-mmlab/mmdetection/blob/ecac3a77becc63f23d9f6980b2a36f86acd00a8a/tools/analysis_tools/get_flops.py#L99) (It seems that the downstream repos have to call this function like the linked line.), not change has to be performed on the existing repos to fit this modification. 

## Use cases (Optional)

Not applicable.

## Checklist

- [x] Pre-commit or other linting tools are used to fix the potential lint issues.
- [x] The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [x] If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
- [x] The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/mmengine/analysis/print_helper.py
+++ b/mmengine/analysis/print_helper.py
@@ -3,7 +3,6 @@
 # Modified from
 # https://github.com/facebookresearch/fvcore/blob/main/fvcore/nn/print_model_statistics.py
 
-import warnings
 from collections import defaultdict
 from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
 
@@ -697,12 +696,10 @@ def get_model_complexity_info(
     Returns:
         dict: The complexity information of the model.
     """
-    assert (input_shape is not None) or (
-        inputs
-        is not None), '"input_shape" and "inputs" cannot both be "None".'
-    if (input_shape is not None) and (inputs is not None):
-        warnings.warn('Both "input_shape" and "inputs" are provided.\
-            "inputs" will be used.')
+    if input_shape is None and inputs is None:
+        raise ValueError('One of "input_shape" and "inputs" should be set.')
+    elif input_shape is not None and inputs is not None:
+        raise ValueError('"input_shape" and "inputs" cannot be both set.')
 
     if inputs is None:
         inputs = (torch.randn(1, *input_shape), )
--- a/mmengine/analysis/print_helper.py
+++ b/mmengine/analysis/print_helper.py
@@ -3,6 +3,7 @@
 # Modified from
 # https://github.com/facebookresearch/fvcore/blob/main/fvcore/nn/print_model_statistics.py
 
+import warnings
 from collections import defaultdict
 from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
 
@@ -675,7 +676,7 @@ def complexity_stats_table(
 
 def get_model_complexity_info(
     model: nn.Module,
-    input_shape: tuple,
+    input_shape: tuple = None,
     inputs: Optional[torch.Tensor] = None,
     show_table: bool = True,
     show_arch: bool = True,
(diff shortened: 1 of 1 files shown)


mmengine/analysis/print_helper.py
```diff
@@ -696,6 +697,13 @@ def get_model_complexity_info(
     Returns:
         dict: The complexity information of the model.
     """
+    assert (input_shape is not None) or (
+        inputs
+        is not None), '"input_shape" and "inputs" cannot both be "None".'
+    if (input_shape is not None) and (inputs is not None):
+        warnings.warn('Both "input_shape" and "inputs" are provided.\
+            "inputs" will be used.')
+
```
zhouzaida, line 706:
```suggestion
    if (input_shape is None) ^ (inputs is None):
        raise ValueError(
            'Only one of input_shape or inputs should be set.')
```

mmengine/analysis/print_helper.py
```diff
@@ -3,6 +3,7 @@
 # Modified from
 # https://github.com/facebookresearch/fvcore/blob/main/fvcore/nn/print_model_statistics.py
 
+import warnings
```
zhouzaida, line 6:
```suggestion
```

mmengine/analysis/print_helper.py
zhouzaida: Directly throwing errors are OK when they are both set.

mmengine/analysis/print_helper.py
```diff
@@ -696,6 +697,13 @@ def get_model_complexity_info(
     Returns:
         dict: The complexity information of the model.
     """
+    assert (input_shape is not None) or (
+        inputs
+        is not None), '"input_shape" and "inputs" cannot both be "None".'
+    if (input_shape is not None) and (inputs is not None):
+        warnings.warn('Both "input_shape" and "inputs" are provided.\
+            "inputs" will be used.')
```
zhouzaida, line 705:
```suggestion
    if input_shape is None and inputs is None:
        raise ValueError(
            'One of the input_shape and inputs should be set.')
    elif input_shape is not None and inputs is not None:
        raise ValueError(
            'input_shape and inputs can not be both set.')
```

mmengine/analysis/print_helper.py
zhouzaida, line 6:
```suggestion
```

mmengine/analysis/print_helper.py
sjiang95: About the error message, how about

```python
if input_shape is None and inputs is None:
    raise ValueError(
        'One of input_shape and inputs should be set.')
# The rest suggestion is fine
```

When neither "input_shape" nor "inputs" is provided, the error message should guide the user to provide one of them. The expression "Only one of input_shape and inputs can be set." is suitable for the scenario that one sets both "input_shape" and "inputs".

Furthermore, I suggest using double quotes as delimiter for "input_shape" and "inputs". That is

```python
if input_shape is None and inputs is None:
    raise ValueError(
        'One of "input_shape" and "inputs" should be set.')
# The rest suggestion is fine
```

as so in https://github.com/open-mmlab/mmengine/blob/1c67f9eb22c5528266c2748d146429c2c7336e19/mmengine/fileio/io.py#L836-L841

zhouzaida: I completely agree. 👍

zhouzaida (approved):

PR #1071, opened 2023-04-12
[Feature] Support Pure Python style Configuration File

HAOCHENYE:

### Add Pure Python style Configuration File

Current pure text style configuration files can satisfy most of our development needs and some module aliases can greatly simplify the configuration files (e.g. `ResNet` can refer to `mmcls.models.ResNet`). However, there are also some disadvantages:

1. In the configuration file, the `type` field is specified by a string, and IDE cannot directly jump to the corresponding class definition, which is not conducive to code reading and jumping.
2. The inheritance of configuration files is also specified by a string, and IDE cannot directly jump to the inherited file. When the inheritance structure of the configuration file is complex, it is not conducive to reading and jumping of the configuration file.
3. The inheritance rules are relatively implicit, and beginners find it difficult to understand how the configuration file merges variables with the same fields and derives special syntax such as `_delete_`, resulting in a higher learning cost.
4. It is easy for users to forget to register the module and cause `module not found` errors.
5. In the yet-to-be-mentioned cross-codebase inheritance, the introduction of the scope makes the inheritance rules of the configuration file more complicated, and beginners find it difficult to understand.

In summary, although pure text style configuration files can provide the same syntax rules for `python`, `json`, and `yaml` format configurations, when the configuration files become complex, pure text style configuration files will appear inadequate. Therefore, we provide a pure Python style configuration file, i.e., the `lazy import` mode, which can fully utilize Python's syntax rules to solve the above problems. At the same time, the pure Python style configuration file also supports exporting to `json` and `yaml` formats.

#### Basic Syntax

Simply describe the syntax difference between python style config and pure text style config

##### Module Construction

We use a simple example to compare pure Python style and pure text style configuration files:

1. Registration for pure Python Style and current pure text style:

  Pure Python style

  ```python
  # No need for registration
  ```
  Pure text style
  ```python
  from torch.optim import SGD
  from mmengine.registry import OPTIMIZERS
  OPTIMIZERS.register_module(module=SGD, name='SGD')
  ```

2. Configuration file writing for pure Python style and current pure text style:

  Pure Python style

  ```python
  # Configuration file writing
  from torch.optim import SGD
  optimizer = dict(type=SGD, lr=0.1)
  ```

  Pure text style

  ```python
  # Configuration file writing
  optimizer = dict(type='SGD', lr=0.1)
  ```

3. Module construction for pure Python style and current pure text style:

  The same for pure Python style and pure text style

  ```python
  import torch.nn as nn
  from mmengine.registry import OPTIMIZERS
  cfg = Config.fromfile('optimizer.py')
  model = nn.Conv2d(1, 1, 1)
  cfg.optimizer.params = model.parameters()
  optimizer = OPTIMIZERS.build(cfg.optimizer)
  ```

From the above example, we can see that the difference between pure Python style and pure text style configuration files is:

1. Pure Python style configuration files do not require module registration.
2. In pure Python style configuration files, the `type` field is no longer a string but directly refers to the module. Correspondingly, import syntax needs to be added in the configuration file.

It should be noted that the OpenMMLab series algorithm library still retains the registration process when adding modules. When users build their own projects based on MMEngine, if they use pure Python style configuration files, registration is not required. You may wonder that if you are not in an environment with torch installed, you cannot parse the sample configuration file. Can this configuration file still be called a configuration file? Don't worry, we will explain this part later.

##### Inheritance

The inheritance syntax of pure Python style configuration files is slightly different:

Pure Python Style:

```python
_base_ = [./optimizer.py]
```

```python
if '_base_':
    from .optimizer import *
```

Pure Python style configuration files use import syntax to achieve inheritance. The advantage of doing this is that we can directly jump to the inherited configuration file for easy reading and jumping. The variable inheritance rule (add, delete, change, and search) is completely aligned with Python syntax. For example, if I want to modify the learning rate of the optimizer in the base configuration file:

```python
if '_base_':
    from .optimizer import *

# optimizer is a variable defined in the base configuration file
optimizer.update(
    lr=0.01,
)
```

Of course, if you are already accustomed to the inheritance rules of pure text style configuration files and the variable is of the `dict` type in the `_base_` configuration file, you can also use merge syntax to achieve the same inheritance rule as pure text style configuration files:

```python
if '_base_':
    from .optimizer import *

# optimizer is a variable defined in the base configuration file
optimizer.merge(
    _delete_=True,
    lr=0.01,
    type='SGD'
)

# The equivalent Python style writing is as follows, completely consistent with Python's import rules
# optimizer = dict(
#     lr=0.01,
#     type='SGD'
# )
```

Compared with pure text style configuration files, the inheritance rule of pure Python style configuration files is completely aligned with the import syntax of Python, which is easier to understand and supports jumping between configuration files. You may wonder since both inheritance and module imports use import syntax, why do we need an `if '_base_'` statement for inheriting configuration files? On the one hand, this can improve the readability of configuration files, making inherited configuration files more prominent. On the other hand, it is also restricted by the rules of lazy_import, which will be explained later.

#### What is Lazy Import

You may find that pure Python style configuration files seem to organize configuration files using pure Python syntax. Then, I do not need configuration classes, and I could just import configuration files using Python syntax. If you have such a feeling, then it is worth celebrating because this is exactly the effect we want.

As mentioned earlier, parsing configuration files requires dependencies on third-party libraries referenced in the configuration files. This is actually a very unreasonable thing. For example, if I trained a model based on MMagic and wanted to deploy it with the onnxruntime backend of MMDeploy. Due to the lack of torch in the deployment environment, and torch is needed in the configuration file parsing process, this makes it inconvenient for me to directly use the configuration file of MMagic as the deployment configuration. To solve this problem, we introduced the concept of lazy_import.

It is a complex task to discuss the specific implementation of lazy_import, so here we only briefly introduce its function. The core idea of lazy_import is to delay the execution of the import statement in the configuration file until the configuration file is parsed, so that the dependency problem caused by the import statement in the configuration file can be avoided. During the configuration file parsing process, the equivalent code executed by the Python interpreter is as follows:

Original configuration file:

```python
from torch.optim import SGD

optimizer = dict(type=SGD)
```

Code actually executed by the python interpreter through the configuration class:

```python
lazy_obj = LazyObject('torch.optim', 'SGD')

optimizer = dict(type=lazy_obj)
```

As an internal type of the `Config` module, the `LazyObject` cannot be accessed directly by users. When accessing the type field, it will undergo a series of conversions to convert `LazyObject` into the actual `torch.optim.SGD` type. In this way, parsing the configuration file will not trigger the import of third-party libraries, while users can still access the types of third-party libraries normally when using the configuration file.

To access the internal type of `LazyObject`, you can use the `Config.to_dict` interface:

```python
cfg = Config.fromfile('optimizer.py').to_dict()
print(type(cfg['optimizer']['type']))
# mmengine.config.lazy.LazyObject
```

At this point, the type accessed is the `LazyObject` type.

However, we cannot adopt the lazy import strategy for the inheritance (import) of base files since we need the configuration file parsed to include the fields defined in the base configuration file, and we need to trigger the import really. Therefore, we have added a restriction on importing base files, which must be imported in the `if '_base_'` code block.

#### Limitations

1. Functions and classes cannot be defined in the configuration file.
2. The configuration file name must comply with the naming convention of Python modules, which can only contain letters, numbers, and underscores, and cannot start with a number.
3. When importing variables from the base configuration file, such as `from ._base_.alpha import beta`, the `alpha` here must be the module (module) name, i.e., a Python file, rather than the package (package) name containing `__init__.py`.
4. Importing multiple variables simultaneously in an absolute import statement, such as `import torch, numpy, os`, is not supported. Multiple import statements need to be used instead, such as `import torch; import numpy; import os`.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

https://github.com/open-mmlab/mmdetection/pull/10366
https://github.com/open-mmlab/mmyolo/pull/787
https://github.com/open-mmlab/mmrazor/pull/539
https://github.com/open-mmlab/mmpose/pull/2390
https://github.com/open-mmlab/mmpretrain/pull/1567

## Checklist


mmengine/config/lazy.py
```diff
@@ -0,0 +1,225 @@
...
+        imported (str, optional): The imported module name. Defaults to None.
+    """
+
+    def __init__(self,
+                 module: Union[str, list, tuple],
+                 imported: Optional[str] = None,
+                 location=None):
+        if (not isinstance(module, str) and not is_seq_of(module, str)):
+            raise TypeError('module should be `str`, `list`, `tuple` or '
+                            f'None, but got {type(module)}, this might be '
```
zhouzaida, line 45: Can the module be None?

mmengine/config/lazy.py
```diff
@@ -0,0 +1,225 @@
...
+    Args:
+        module (str or list or tuple): The module name to be imported.
+        imported (str, optional): The imported module name. Defaults to None.
+    """
+
+    def __init__(self,
+                 module: Union[str, list, tuple],
+                 imported: Optional[str] = None,
+                 location=None):
+        if (not isinstance(module, str) and not is_seq_of(module, str)):
```
zhouzaida, line 43:
```suggestion
        if not isinstance(module, str) and not is_seq_of(module, str):
```

mmengine/config/lazy.py
```diff
@@ -0,0 +1,225 @@
...
+
+    Args:
+        module (str or list or tuple): The module name to be imported.
+        imported (str, optional): The imported module name. Defaults to None.
+    """
+
+    def __init__(self,
+                 module: Union[str, list, tuple],
+                 imported: Optional[str] = None,
+                 location=None):
```
zhouzaida, line 42: type hint and docstring for it.

mmengine/model/averaged_model.py
```diff
@@ -258,5 +258,6 @@ def avg_func(self, averaged_param: Tensor, source_param: Tensor,
             steps (int): The number of times the parameters have been
                 updated.
         """
-        momentum = max(self.momentum, self.gamma / (self.gamma + self.steps))
+        momentum = max(self.momentum,
+                       self.gamma / (self.gamma + self.steps.item()))
```
zhouzaida, line 262: Is it necessary to call `item()`?

mmengine/config/utils.py
```diff
@@ -144,3 +149,282 @@ def visit_Assign(self, node):
...
+    if ('site-package' in origin_path
+            or not origin_path.startswith(PYTHON_ROOT_DIR)):
+        return False
+    else:
+        return True
+
+
+class Transform(ast.NodeTransformer):
+    """Convert the import syntax to the assignment of
+    ``mmengine.config.LazyObject`` and preload the base variable before parsing
```
zhouzaida, line 181:
```suggestion
    :class:`mmengine.config.LazyObject` and preload the base variable before parsing
```

mmengine/config/__init__.py
```diff
@@ -1,4 +1,8 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 from .config import Config, ConfigDict, DictAction
+from .lazy import LazyAttr, LazyObject
+from .utils import Transform
```
zhouzaida, line 4: Users do not need to use those classes so they can be removed from this `__init__` file.

mmengine/config/utils.py
```diff
@@ -144,3 +149,282 @@ def visit_Assign(self, node):
...
+                lineno = name.lineno
+            else:
+                lineno = node.lineno
+            if name == '*':
+                # TODO If user import * from a non-config module, it should
+                # fallback to import the real module and raise a warning to
+                # remind user the real module will be imported which will slows
+                # donwn the parsing speed.
+                raise RuntimeError(
+                    'You cannot import * from a non-config module!')
```
zhouzaida, line 325: need to refine this error message

mmengine/config/config.py
```diff
@@ -42,14 +44,54 @@ class ConfigDict(Dict):
 
     The Config class would transform the nested fields (dictionary-like fields)
     in config file into ``ConfigDict``.
+
+    If the class attribute``lazy``  is ``False``, users will get the
+    object built by ``LazyObject`` or ``LazyAttr``, otherwise users will get
+    the ``LazyObject`` or ``LazyAttr`` itself.
```
zhouzaida, line 50:
```suggestion
    If the class attribute ``lazy``  is ``False``, users will get the
    object built by ``LazyObject`` or ``LazyAttr``, otherwise users will get
    the ``LazyObject`` or ``LazyAttr`` itself.
```

mmengine/config/config.py
```diff
@@ -42,14 +44,54 @@ class ConfigDict(Dict):
...
+
+    The ``lazy`` should be set to ``True`` to avoid building the imported
+    object during configuration parsing, and it should be set to False outside
+    the Config to ensure that users do not experience the ``LazyObject``.
     """
+    # Lazy should be False when parsing configs, otherwise the real object
+    # will be imported during parsing.
+    # Lazy should be True after parsing the configs, users will get the
+    # real object and will not perceive the existence of LazyObject
```
zhouzaida, line 59: Those lines can be removed because they are repeated with the docstring.

mmengine/model/averaged_model.py
HAOCHENYE: I got an error in some conditions but I cannot remember how could it happened :rofl:. But I think it could be more proper to compare the builtin variable with the builtin function `max`

mmengine/config/utils.py
```diff
@@ -144,3 +149,282 @@ def visit_Assign(self, node):
...
+        assert len(alias_list) == 1, (
+            'Does not support import multiple modules in one line')
+        # TODO Support multiline import
+        alias = alias_list[0]
+        if alias.asname is not None:
+            self.imported_obj.add(alias.asname)
+            return ast.parse(  # type: ignore
+                f'{alias.asname} = LazyObject('
+                rf'"{alias.name}",'
+                rf'location="{self.filename}: line, {node.lineno}")').body[0]
```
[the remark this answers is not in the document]
HAOCHENYE: `re` cannot parse the windows path like `D:\\xx\xx` with out the prefix `r`

docs/zh_cn/api/config.rst
```diff
@@ -14,3 +14,6 @@ mmengine.config
    Config
    ConfigDict
    DictAction
+   Transform
+   LazyObject
+   LazyAttr
```
zhouzaida, line 19:
```suggestion
```

docs/en/api/config.rst
```diff
@@ -14,3 +14,6 @@ mmengine.config
    Config
    ConfigDict
    DictAction
+   Transform
+   LazyObject
+   LazyAttr
```
zhouzaida, line 19:
```suggestion
```

mmengine/config/lazy.py
```diff
@@ -0,0 +1,225 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import importlib
+from typing import Any, Optional, Union
+
+from mmengine.utils import is_seq_of
+
+
+class LazyObject:
+    """LazyObject is used to lazy initialize the imported module during parsing
+    the configuration file.
```
zhouzaida, line 10:
```suggestion
    """LazyObject is used to lazily initialize the imported module during parsing
    the configuration file.
```

mmengine/config/lazy.py
```diff
@@ -0,0 +1,225 @@
...
+    Examples:
+        >>> # import torch.nn as nn
+        >>> nn = lazyObject('torch.nn')
+        >>> # from mmdet.models import RetinaNet
+        >>> RetinaNet = lazyObject(['mmdet.models', 'RetinaNet'])
+        >>> # import mmcls.models; import mmcls.datasets; import mmcls
+        >>> mmcls = lazyObject(['mmcls', 'mmcls.datasets', 'mmcls.models'])
+
+    ``LazyObject`` records all module information and will be further
+    referenced by configuration file.
```
zhouzaida, line 32:
```suggestion
    ``LazyObject`` records all module information and will be further
    referenced by the configuration file.
```

mmengine/config/utils.py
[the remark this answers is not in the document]
zhouzaida: So why do not add `r` in https://github.com/open-mmlab/mmengine/pull/1071/files#diff-b7e0842403cd680e4a8deb5604791fd694da752a4cffc9fb4b5d3ec29824e982R330?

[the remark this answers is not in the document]
zhouzaida: And why does the delimiter is `:` here but `,` in https://github.com/open-mmlab/mmengine/pull/1071/files#diff-b7e0842403cd680e4a8deb5604791fd694da752a4cffc9fb4b5d3ec29824e982R330?

mmengine/config/utils.py
```diff
@@ -144,3 +149,282 @@ def visit_Assign(self, node):
...
+                    base_key = name.asname
+                else:
+                    base_key = name.name
+                self.global_dict[base_key] = self.base_dict[module][name.name]
+            return None
+
+        # TODO: Support lazyimport module from relative path
+        nodes: List[ast.Assign] = []
+        for name in node.names:
+            # `ast.alias`` have lineno attr after Python 3.10,
```
zhouzaida, line 314:
```suggestion
            # `ast.alias` have lineno attr after Python 3.10,
```

mmengine/config/lazy.py
```diff
@@ -0,0 +1,225 @@
...
+                        f'Cannot import {self._imported} '
+                        f'from {self._module} in {self.location}')
+                else:
+                    raise ImportError(f'Cannot import {self._module} '
+                                      f'in {self.location}')
+            except Exception as e:
+                raise type(e)(
+                    f'{e} in {self.location}, if you have checked the '
+                    'correctness of your syntax, please report this bug '
+                    'to https://github.com/open-mmlab/mmengine/issues')
```
zhouzaida, line 82:
```suggestion
            try:
                module = importlib.import_module(self._module)
            except ImportError:
                raise ImportError(f'Cannot import {self._module} '
                                  f'in {self.location}')
                                  
            if self._imported is not None:
                if hasattr(module, self._imported):
                    module = getattr(module, self._imported)
                else:
                    raise ImportError(
                        f'Cannot import {self._imported} '
                        f'from {self._module} in {self.location}')
```

mmengine/config/utils.py
```diff
@@ -144,3 +149,282 @@ def visit_Assign(self, node):
...
+        return False
+    origin_path = osp.abspath(origin_path)
+    if ('site-package' in origin_path
+            or not origin_path.startswith(PYTHON_ROOT_DIR)):
+        return False
+    else:
+        return True
+
+
+class Transform(ast.NodeTransformer):
```
zhouzaida, line 179:
```suggestion
class ImportTransformer(ast.NodeTransformer):
```

mmengine/config/utils.py
```diff
@@ -144,3 +149,282 @@ def visit_Assign(self, node):
...
+                code = f'{name.name} = LazyObject("{module}", "{name.name}", "{self.filename}, line {lineno}")'  # noqa: E501
+                self.imported_obj.add(name.name)
+            try:
+                nodes.append(ast.parse(code).body[0])  # type: ignore
+            except Exception as e:
+                raise ImportError(
+                    f'Cannot import {name} from {module}',
+                    '1. Cannot import * from 3rd party lib in the config file',
+                    '2. Please check if the module is a base config which '
+                    'should be added to `_base_`',
```
zhouzaida, line 345: Please add ` ` or `\n` between lines in the error message to make it better

mmengine/config/lazy.py
```diff
@@ -0,0 +1,225 @@
...
+        self.location = location
+
+    def build(self) -> Any:
+        """Return imported object.
+
+        Returns:
+            Any: Imported object
+        """
+        if isinstance(self._module, str):
+            # For import xxx.xxx as xxx or from xxx.xxx import xxx
```
zhouzaida, line 65:
```suggestion
            # For import xxx.yyy as zzz or from xxx.yyy import zzz
```

mmengine/config/lazy.py
zhouzaida: We need to make the `try-except` explicit.

mmengine/config/config.py
```diff
@@ -109,6 +295,8 @@ class Config:
         filename (str or Path, optional): Name of config file.
             Defaults to None.
 
+    Here is a simple examples:
```
zhouzaida, line 298:
```suggestion
    Here is a simple example:
```

mmengine/config/config.py
```diff
@@ -143,7 +335,9 @@ def __init__(self,
             if key in RESERVED_KEYS:
                 raise KeyError(f'{key} is reserved for config file')
 
-        super().__setattr__('_cfg_dict', ConfigDict(cfg_dict))
+        if not isinstance(cfg_dict, ConfigDict):
+            cfg_dict = ConfigDict(cfg_dict)
```
zhouzaida, line 339: Please also update the type hint.

mmengine/config/config.py
```diff
@@ -161,40 +355,59 @@ def __init__(self,
...
         """Build a Config instance from config file.
 
         Args:
             filename (str or Path): Name of config file.
             use_predefined_variables (bool, optional): Whether to use
                 predefined variables. Defaults to True.
             import_custom_modules (bool, optional): Whether to support
-                importing custom modules in config. Defaults to True.
+                importing custom modules in config. Defaults to None. If it
+                is None, it will be deduced by the config file style.
```
zhouzaida, line 368: typo?

mmengine/config/config.py
```diff
@@ -1020,6 +1480,49 @@ def merge_from_dict(self,
             Config._merge_a_into_b(
                 option_cfg_dict, cfg_dict, allow_list_keys=allow_list_keys))
 
+    @staticmethod
+    def _is_lazy_import(filename):
+        with open(filename) as f:
```
zhouzaida, line 1485:
```suggestion
        with open(filename, encoding='utf-8') as f:
```

mmengine/config/config.py
```diff
@@ -42,14 +44,54 @@ class ConfigDict(Dict):
...
+    object during configuration parsing, and it should be set to False outside
+    the Config to ensure that users do not experience the ``LazyObject``.
     """
+    # Lazy should be False when parsing configs, otherwise the real object
+    # will be imported during parsing.
+    # Lazy should be True after parsing the configs, users will get the
+    # real object and will not perceive the existence of LazyObject
+    lazy = False
+
+    def __init__(__self, *args, **kwargs):
```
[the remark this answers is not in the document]
HAOCHENYE: Just follow the manner of addict.

mmengine/config/config.py
```diff
@@ -1020,6 +1480,49 @@ def merge_from_dict(self,
             Config._merge_a_into_b(
                 option_cfg_dict, cfg_dict, allow_list_keys=allow_list_keys))
 
+    @staticmethod
+    def _is_lazy_import(filename):
+        with open(filename) as f:
+            codes_str = f.read()
+            codes = ast.parse(codes_str)
```
zhouzaida, line 1487:
```suggestion
            nodes = ast.parse(codes_str)
```

mmengine/config/config.py
```diff
@@ -1020,6 +1480,49 @@ def merge_from_dict(self,
...
+        for node in ast.walk(codes):
```
zhouzaida, line 1488:
```suggestion
            nodes = ast.parse(codes_str)
        for node in ast.walk(nodes):
```

mmengine/config/config.py
```diff
@@ -161,40 +355,59 @@ def __init__(self,
...
+            # enable lazy import when parsing the config
+            ConfigDict.lazy = True
+            cfg_dict, imported_names = Config._parse_lazy_import(filename)
+            for key, value in list(cfg_dict.to_dict().items()):
+                if isinstance(value, (types.FunctionType, types.ModuleType)):
+                    cfg_dict.pop(key)
+            # disable lazy import to get the real type.
+            ConfigDict.lazy = False
+            cfg = Config(cfg_dict, filename=filename)
+            object.__setattr__(cfg, '_imported_names', imported_names)
```
zhouzaida, line 409:
```suggestion
            super().__setattr__(cfg, '_imported_names', imported_names)
```

mmengine/config/config.py
```diff
@@ -232,6 +445,61 @@ def fromstring(cfg_str: str, file_format: str) -> 'Config':
         os.remove(temp_file.name)  # manually delete the temporary file
         return cfg
 
+    @staticmethod
+    def _get_base_modules(codes: list) -> list:
+        """Get base module name from python file.
+
+        Args:
+            filename (str): Name of python config file.
+
```
zhouzaida, line 454: The docstring is not consistent with the method signature and function.

mmengine/config/config.py
```diff
@@ -573,6 +841,174 @@ def _file2dict(
...
+            ConfigDict: Converted dict.
+        """
+        # Only the outer dict with key `type` should have the key `_scope_`.
+        if isinstance(cfg, dict):
+            if isinstance(cfg, ConfigDict):
+                # Use to_dict to avoid build lazy object.
+                cfg = cfg.to_dict()
+            cfg_dict = ConfigDict()
+            for key, value in cfg.items():
+                cfg_dict[key] = Config._dict_to_config_dict_lazy(value, )
```
zhouzaida, line 1005:
```suggestion
                cfg_dict[key] = Config._dict_to_config_dict_lazy(value)
```

mmengine/config/lazy.py
HAOCHENYE: Both `from ... import xxx` and `import xxx` require that the `xxx` after the `import`. Therefore module could not be `None`

mmengine/config/config.py
HAOCHENYE: We can use `super` here since it is a staticmethod

mmengine/config/lazy.py
```diff
@@ -0,0 +1,217 @@
...
+        if self.location is not None:
+            location = self.location.split(', line')
+        else:
+            location = self.location
+        return LazyAttr(name, self, location)
+
+    def __str__(self) -> str:
+        if self._imported is not None:
+            return str(self._imported)
+        return self.module
```
zhouzaida, line 114:
```suggestion
        return self._module
```

mmengine/config/lazy.py
```diff
@@ -0,0 +1,217 @@
...
+        module (str or list or tuple): The module name to be imported.
+        imported (str, optional): The imported module name. Defaults to None.
+        location (str, optional): The filename and line number of the imported
+            module statement happened.
+    """
+
+    def __init__(self,
+                 module: Union[str, list, tuple],
+                 imported: Optional[str] = None,
+                 location: str = None):
```
zhouzaida, line 44:
```suggestion
                 location: Optional[str] = None):
```

mmengine/config/utils.py
```diff
@@ -144,3 +149,297 @@ def visit_Assign(self, node):
...
+    ) -> Optional[Union[List[ast.Assign], ast.ImportFrom]]:
+        """Hack the ``from ... import ...`` syntax and update the global_dict.
+
+        Examples:
+            >>> from mmdet.models import RetinaNet
+
+        Will be parsed as:
+
+        Examples:
+            >>> RetinaNet = lazyObject(['mmdet.models', 'RetinaNet'])
```
zhouzaida, line 287: Should it be `RetinaNet = lazyObject('mmdet.models', 'RetinaNet')`?

mmengine/config/lazy.py
```diff
@@ -0,0 +1,217 @@
...
+        >>> import mmcls.datasets
+        >>> import mmcls
+
+    Will be parsed as:
+
+    Examples:
+        >>> # import torch.nn as nn
+        >>> nn = lazyObject('torch.nn')
+        >>> # from mmdet.models import RetinaNet
+        >>> RetinaNet = lazyObject(['mmdet.models', 'RetinaNet'])
```
zhouzaida, line 27: Should it be `RetinaNet = lazyObject('mmdet.models', 'RetinaNet')`?

mmengine/config/lazy.py
```diff
@@ -0,0 +1,217 @@
...
+        """Return the attribute of the imported object.
+
+        Returns:
+            Any: attribute of the imported object.
+        """
+        try:
+            return getattr(self.source.build(), self.name)
+        except AttributeError:
+            raise ImportError(
+                f'Cannot import f{self.module} in {self.location}')
```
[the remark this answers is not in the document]
zhouzaida: The try-except block should be split into two parts. The first part is to build the module. The second is to get the `self.name`. And their error message should be different.

mmengine/config/config.py
```diff
@@ -58,6 +97,150 @@ def __getattr__(self, name):
...
+        """
+        return self.build_lazy(super().get(key, default))
+
+    def pop(self, key, default=None):
+        """Pop the value of the key. If class attribute ``lazy`` is True, the
+        LazyObject will be built and returned.
+
+        Args:
+            key (str): The key.
+            default (any, optional): _description_. Defaults to None.
```
zhouzaida, line 137:
```suggestion
            default (any, optional): The default value. Defaults to None.
```

mmengine/config/config.py
```diff
@@ -42,14 +45,50 @@ class ConfigDict(Dict):
...
+            elif isinstance(arg, dict):
+                for key, val in arg.items():
+                    __self[key] = __self._hook(val)
+            elif isinstance(arg, tuple) and (not isinstance(arg[0], tuple)):
+                __self[arg[0]] = __self._hook(arg[1])
+            else:
+                for key, val in iter(arg):
+                    __self[key] = __self._hook(val)
+
+        for key, val in kwargs.items():
```
zhouzaida, line 81: Can `kwargs` be a `ConfigDict`? If it can, should we also use `dict(kwargs).item()` as https://github.com/open-mmlab/mmengine/pull/1071/files#diff-aa8376fc15b5b802ef25189d351d2580a2b77c41f3962c87fbef546cafc2c2d7R155

mmengine/config/config.py
```diff
@@ -58,6 +97,150 @@ def __getattr__(self, name):
...
+                return {
+                    key: _to_dict(value)
+                    for key, value in Dict.items(data)
+                }
+            elif isinstance(data, dict):
+                return {key: _to_dict(value) for key, value in data.items()}
+            elif isinstance(data, (list, tuple)):
+                return type(data)(_to_dict(item) for item in data)
+            else:
+                return data
```
zhouzaida, line 240: It seems like the `_to_dict` is not necessary. Just call `to_dict` enough.

mmengine/config/config.py
```diff
@@ -232,6 +450,74 @@ def fromstring(cfg_str: str, file_format: str) -> 'Config':
         os.remove(temp_file.name)  # manually delete the temporary file
         return cfg
 
+    @staticmethod
+    def _get_base_modules(codes: list) -> list:
+        """Get base module name from parsed code.
+
+        Args:
+            codes (str): Parsed code of the config file.
```
zhouzaida, line 458:
```suggestion
            codes (list): Parsed code of the config file.
```

mmengine/config/config.py
```diff
@@ -232,6 +450,74 @@ def fromstring(cfg_str: str, file_format: str) -> 'Config':
...
+
+            Returns:
+                list: Name of base modules.
+            """
+            base_modules = []
+            for inst in ifcodes:
+                assert isinstance(inst, ast.ImportFrom), (
+                    'Only support import from statement in _base_')
+                assert inst.module is not None, (
+                    'You must import base file by from xxx import yyy')
```
zhouzaida, line 478: refine

mmengine/config/config.py
```diff
@@ -232,6 +450,74 @@ def fromstring(cfg_str: str, file_format: str) -> 'Config':
...
+                base_modules.append(inst.level * '.' + inst.module)
+            return base_modules
+
+        for idx, inst in enumerate(codes):
```
zhouzaida, line 482: What is the meaning of `inst`?

mmengine/config/config.py
```diff
@@ -1020,6 +1515,53 @@ def merge_from_dict(self,
...
+                    continue
+                if not isinstance(node.module, str):
+                    continue
+                # non-builtin module -> lazy_import
+                if not _is_builtin_module(node.module):
+                    return True
+            if isinstance(node, ast.Import):
+                for name in node.names:
+                    if not _is_builtin_module(name.name):
+                        return True
```
zhouzaida, line 1550:
```suggestion
                for alias_node in node.names:
                    if not _is_builtin_module(alias_node.name):
                        return True
```

mmengine/config/config.py
HAOCHENYE: `inst` means the instruction, which is an `ast.Node` object. I just follow the manner of PyTorch Dynamo, which names the bytecode as `inst`

mmengine/config/lazy.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def build(self) -> Any:
+        """Return the attribute of the imported object.
+
+        Returns:
+            Any: attribute of the imported object.
+        """
+        source = self.source.build()
+        try:
+            return getattr(source, self.name)
```
zhouzaida, line 208:
```suggestion
        module = self.source.build()
        try:
            return getattr(module, self.name)
```

mmengine/config/lazy.py
```diff
@@ -0,0 +1,218 @@
...
+        # b.c.d -> self._module of d = `a.b.c`
+        if isinstance(self.source, LazyObject):
+            if isinstance(self.source._module, str):
+                # In this case, the source code of LazyObject could be one of
+                # the following:
+                # 1. import xxx.yyy as zzz
+                # 2. from xxx.yyy import zzz
+
+                # The equivalent code of LazyObject is:
+                # zzz = LazyObject('')
```
zhouzaida, line 159:
```suggestion
                # 1. import xxx.yyy as zzz
                # 2. from xxx.yyy import zzz

                # The equivalent code of LazyObject is:
                # 1. zzz = LazyObject('xxx.yyy')
                # 2. zzz = LazyObject('xxx.yyy', 'zzz')
```

mmengine/config/lazy.py
```diff
@@ -0,0 +1,218 @@
...
+    def __init__(self,
+                 name: str,
+                 source: Union['LazyObject', 'LazyAttr'],
+                 location=None):
+        self.name = name
+        self.source: Union[LazyAttr, LazyObject] = source
+        # What is self._module meaning:
+        # import a.b as b
+        # b.c -> self._module of c is `a.b`
+        # b.c.d -> self._module of d = `a.b.c`
```
zhouzaida, line 150:
```suggestion
        # import a.b as c
        # c.d -> self._module of d is `a.b`
        # c.d.e -> self._module of e is `a.b.d`
```

mmengine/config/utils.py
```diff
@@ -144,3 +149,299 @@ def visit_Assign(self, node):
...
         else:
             return node
+
+
+def _is_builtin_module(module_name: str) -> bool:
+    """Check if a module is a built-in module.
+
+    Arg:
+        module_name: name of module.
+    """
```
zhouzaida, line 159: `sys` has an attribute `builtin_module_names` so can we re-use it to simplify the helper function?

mmengine/config/utils.py
```diff
@@ -144,3 +149,299 @@ def visit_Assign(self, node):
...
+    A very intuitive idea for solving this problem is to import the module
+    corresponding to the "type" field using the ``import`` syntax. Similarly,
+    we can also ``import`` base file.
+
+    However, this approach has a significant drawback. We need to trigger the
+    import logic to parse the configuration file, which can be time-consuming.
+    This also implies that we have to download numerous dependencies to parse
+    the configuration file. Yet, there may be instances where we simply want
+    to use string-like fields in the configuration rather than actually
+    importing them.
```
zhouzaida, line 208:
```suggestion
    However, this approach has a significant drawback. We need to trigger the
    import logic to parse the configuration file, which can be time-consuming.
    This also implies that we have to install numerous dependencies to parse
    the configuration file. Yet, there may be instances where we simply want
    to use string-like fields in the configuration rather than actually
    importing them.
```

mmengine/config/utils.py
```diff
@@ -144,3 +149,299 @@ def visit_Assign(self, node):
...
+
+    To avoid this problem, we introduce :class:`mmengine.config.LazyObject` and
+    :class:`mmengine.config.LazyAttr`. Before we proceed with further
+    explanations, you may refer to the documentation of these two modules to
+    gain an understanding of their functionalities.
+
+    Actually, one of the functions of ``Transform`` is to hack the ``import``
```
zhouzaida, line 215:
```suggestion
    Actually, one of the functions of ``ImportTransformer`` is to hack the ``import``
```

mmengine/config/utils.py
```diff
@@ -144,3 +149,299 @@ def visit_Assign(self, node):
...
+    syntax. It will replace the import syntax (exclude import the base files)
+    with the assignment of ``LazyObject``.
+
+    As for the import syntax of the base file, we cannot lazy import it since
+    we're eager to merge the fields of current file and base files. Therefore,
+    another function of the ``Transform`` is to collaborate with
```
zhouzaida, line 221:
```suggestion
    another function of the ``ImportTransformer`` is to collaborate with
```

mmengine/config/utils.py
```diff
@@ -144,3 +149,299 @@ def visit_Assign(self, node):
...
+
+            Examples:
+                >>> global_dict.update(base_dict['.._base_.default_runtime'])  # `import *` means update all data
+                >>> global_dict.update(dataset=base_dict['.._base_.datasets.coco_detection']['dataset'])  # only update `dataset`
+    """  # noqa: E501
+
+    def __init__(self,
+                 global_dict: dict,
+                 base_dict: Optional[dict] = None,
+                 filename=None):
```
zhouzaida, line 258:
```suggestion
                 filename: Optional[str] = None):
```

mmengine/config/utils.py
HAOCHENYE: It seems `sys.builtin_module_name` cannot list all builtin modules:

```bash
In [2]: sys.builtin_module_names
Out[2]: 
('_abc',
 '_ast',
 '_codecs',
 '_collections',
 '_functools',
 '_imp',
 '_io',
 '_locale',
 '_operator',
 '_signal',
 '_sre',
 '_stat',
 '_string',
 '_symtable',
 '_thread',
 '_tracemalloc',
 '_warnings',
 '_weakref',
 'atexit',
 'builtins',
 'errno',
 'faulthandler',
 'gc',
 'itertools',
 'marshal',
 'posix',
 'pwd',
 'sys',
 'time',
 'xxsubtype')
```

As you see, `os` is not in the result.

zhouzaida: got it.

mmengine/config/utils.py
```diff
@@ -144,3 +149,301 @@ def visit_Assign(self, node):
...
+                    return None
+                if alias_node.asname is not None:
+                    base_key = alias_node.asname
+                else:
+                    base_key = alias_node.name
+                self.global_dict[base_key] = self.base_dict[module][
+                    alias_node.name]
+            return None
+
+        # TODO: Support lazyimport module from relative path
```
zhouzaida, line 325:
```suggestion
        # TODO: Support lazily importing modules from a relative path
```

mmengine/config/utils.py
```diff
@@ -144,3 +149,301 @@ def visit_Assign(self, node):
...
+            # `ast.alias` has lineno attr after Python 3.10,
+            if hasattr(alias_node, 'lineno'):
+                lineno = alias_node.lineno
+            else:
+                lineno = node.lineno
+            if alias_node == '*':
+                # TODO If user import * from a non-config module, it should
+                # fallback to import the real module and raise a warning to
+                # remind user the real module will be imported which will slows
+                # donwn the parsing speed.
```
zhouzaida, line 337:
```suggestion
                # TODO: If users import * from a non-config module, it should
                # fallback to import the real module and raise a warning to
                # remind users the real module will be imported which will slow
                # down the parsing speed.
```

mmengine/config/config.py
```diff
@@ -58,6 +97,150 @@ def __getattr__(self, name):
...
+
+        Args:
+            other (dict): Another dictionary.
+        """
+        default = object()
+
+        def _merge_a_into_b(a, b):
+            if isinstance(a, dict):
+                if not isinstance(b, dict):
+                    a.pop('_delete_', None)
```
zhouzaida, line 207:
```suggestion
                    a.pop(DELETE_KEY, None)
```

mmengine/config/config.py
```diff
@@ -58,6 +97,150 @@ def __getattr__(self, name):
...
+                    return a
+                if DELETE_KEY in a:
```
zhouzaida, line 209: Do we need to check `DELETE_KEY`'s value (True or False) before clearing b?

mmengine/config/config.py
```diff
@@ -161,40 +352,67 @@ def __init__(self,
...
-            env_variables=env_variables)
+                cfg_dict, imported_names = Config._parse_lazy_import(filename)
+            except Exception as e:
+                raise e
+            finally:
+                ConfigDict.lazy = False
+            for key, value in list(cfg_dict.to_dict().items()):
+                if isinstance(value, (types.FunctionType, types.ModuleType)):
+                    cfg_dict.pop(key)
+            # disable lazy import to get the real type.
```
zhouzaida, line 412: What is the meaning of the comment?

mmengine/config/config.py
```diff
@@ -232,6 +450,77 @@ def fromstring(cfg_str: str, file_format: str) -> 'Config':
...
+            #     from .._base_.default_runtime import *
+            # ```
+            # The processed code:
+            # ```
+            # from .._base_.default_runtime import *
+            # ```
+            # As you can see, the if statement is removed and the
+            # from ... import statement will be unindent
+            for nested_idx, nested_inst in enumerate(node.body):
+                nodes.insert(idx + nested_idx + 1, nested_inst)
```
zhouzaida, line 519:
```suggestion
            for nested_idx, nested_node in enumerate(node.body):
                nodes.insert(idx + nested_idx + 1, nested_node)
```

mmengine/config/config.py
```diff
@@ -573,6 +880,174 @@ def _file2dict(
...
         return cfg_dict, cfg_text, env_variables
 
+    @staticmethod
+    def _parse_lazy_import(filename: str) -> Tuple[ConfigDict, set]:
+        """Transform file to variables dictionary.
+
+        Args:
+            filename (str): Name of config file.
+            use_predefined_variables (bool, optional): Whether to use
+                predefined variables. Defaults to True.
```
zhouzaida, line 890: clear this unused argument

mmengine/config/config.py
```diff
@@ -573,6 +880,174 @@ def _file2dict(
...
+                    if len(module_list) == 1:
+                        raise RuntimeError(
+                            'The imported configuration file should not be '
+                            f'an independent package {module_list[0]}. Here '
+                            'is an example: '
+                            '"_base_ = mmdet.configs.retinanet_r50_fpn_1x_coco"'  # noqa: E501
+                        )
+                    else:
+                        package = module_list[0]
+                        root_path = get_installed_path(package)
```
zhouzaida, line 972: The package name of `mmseg` is `mmsegmentation` so can the `get_installed_path` returns a correct `root_path` when passing `mmseg` rather than `mmsegmentation` to it?

mmengine/config/config.py
```diff
@@ -573,6 +880,174 @@ def _file2dict(
...
+                base_cfg, _base_imported_names = Config._parse_lazy_import(
+                    module_path)
+                base_imported_names |= _base_imported_names
+                # The base_dict will be:
+                # {
+                #     'mmdet.configs.default_runtime': {...}
+                #     'mmdet.configs.retinanet_r50_fpn_1x_coco': {...}
+                #     ...
+                # }
+                base_dict[base_module] = base_cfg
```
zhouzaida, line 991:
```suggestion
                _base_dict, _base_imported_names = Config._parse_lazy_import(
                    module_path)
                base_imported_names |= _base_imported_names
                # The base_dict will be:
                # {
                #     'mmdet.configs.default_runtime': {...}
                #     'mmdet.configs.retinanet_r50_fpn_1x_coco': {...}
                #     ...
                # }
                base_dict[base_module] = _base_dict
```

mmengine/utils/misc.py
```diff
@@ -500,3 +500,47 @@ def wrapper(*args, **kwargs):
...
+            module_name = f'{module_name}.{part}'
+        except StopIteration:
+            # if obj is a module
+            return module
+        except ImportError:
+            return None
+
+    # module does not exist
+    if module is None:
+        return module
```
zhouzaida, line 536: In what cases would a module be None?

mmengine/utils/misc.py
```diff
@@ -500,3 +500,47 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+
+def locate(obj_name: str):
```
zhouzaida, line 505:
```python
def locate(name: str)
    module_name, obj_name = string.rsplit('.', 1)
    module = importlib.import_module(module_name)
    obj = getattr(module, obj_name)
    return obj
```

HAOCHENYE: Both ways cannot locate the class attribute in one module since it is unknown that which part of the string is a module, and which is the imported object.  

BTW, the name `locate` is borrowed from the `pydoc.locate`

mmengine/utils/misc.py
HAOCHENYE: It is the default value for `getattr`

mmengine/config/config.py
HAOCHENYE: I've also modified the implementation of `get_installed_path`

```bash
/home/yehaochen/anaconda3/envs/py310torch20/lib/python3.10/site-packages/pkg_resources/__init__.py:122: PkgResourcesDeprecationWarning: v2.2.1-2-g1505ef3-master is an invalid version and will not be supported in a future release
  warnings.warn(
Out[8]: '/home/yehaochen/anaconda3/envs/py310torch20/lib/python3.10/site-packages/mmseg'
```

Using `find_spec` can find the install path of `mmseg`

mmengine/config/config.py
HAOCHENYE: It just means disabling the lazy mode of ConfigDict. I've described it the usage of meaning in the `ConfigDict`'s docstring. If it is not enough, I can implement a context manager and add more docstring there.

mmengine/config/config.py
HAOCHENYE: Replace with `a.pop(DELETE_KEY, False)`

mmengine/config/utils.py
HAOCHENYE: Actually, this have been implemented, just remove it.

docs/zh_cn/advanced_tutorials/config.md
```diff
@@ -364,12 +368,445 @@ a=1
...
+cfg = Config.fromfile('optimizer.py').to_dict()
+print(type(cfg['optimizer']['type']))
+# mmengine.config.lazy.LazyObject
+```
+
+此时访问得到的 type 就是 LazyObject 类型。
+
+然而对于 base 文件的继承（导入，import），我们不能够采取 lazy import 的策略，这是因为我们希望解析后的配置文件能够包含 base 配置文件定义的字段，需要真正的触发 import。因此我们对 base 文件的导入加了一层限制，即必须在 `if '_base_'` 的代码块中导入。
+
+关于 LazyObject 的更多细节，可以参考 [Transform](mmengine.config.Transform)、[LazyObject](mmengine.config.LazyObject)， [LazyAttr](mmengine.config.LazyAttr) 的 api 文档。
```
zhouzaida, line 605: API can not be found so they can be removed here or add the link of code.

docs/zh_cn/advanced_tutorials/config.md
```diff
@@ -25,6 +25,10 @@ wget https://raw.githubusercontent.com/open-mmlab/mmengine/main/docs/resources/c
 wget https://raw.githubusercontent.com/open-mmlab/mmengine/main/docs/resources/config/modify_base_var.py
 ```
 
+:::{note}
+如果你已经对配置文件的使用方式有了一定的了解，但是尚且不了解配置文件的 lazy_import 模式，可以跳转到 [纯 python 风格配置文件](#纯-python-风格的配置文件)一节进行阅读
```
zhouzaida, line 29:
```suggestion
如果你已经对配置文件的使用方式有了一定的了解，但是尚且不了解配置文件的 lazy_import 模式，可以跳转到[纯 python 风格配置文件](#纯-python-风格的配置文件)一节进行阅读
```

docs/zh_cn/advanced_tutorials/config.md
```diff
@@ -364,12 +368,447 @@ a=1
...
+            cfg.optimizer.params = model.parameters()
+            optimizer = OPTIMIZERS.build(cfg.optimizer)
+```
+
+从上面的例子可以看出，纯 Python 风格的配置文件和纯文本风格的配置文件的区别在于：
+
+1. 纯 Python 风格的配置文件无需注册模块
+2. 纯 Python 风格的配置文件中，type 字段不再是字符串，而是直接指代模块。相应的配置文件需要多出 import 语法
+
+需要注意的是，OpenMMLab 系列算法库在新增模块时仍会保留注册过程，用户基于 MMEngine 构建自己的项目时，如果使用纯 python 风格的配置文件，则无需注册。看到这你会或许会好奇，这样没有安装 torch 的环境不就没法解析样例配置文件了么，这样的配置文件还叫配置文件么？不要着急，这部分的内容我们会在后面介绍
```
zhouzaida, line 454:
```suggestion
需要注意的是，OpenMMLab 系列算法库在新增模块时仍会保留注册过程，用户基于 MMEngine 构建自己的项目时，如果使用纯 python 风格的配置文件，则无需注册。看到这你会或许会好奇，这样没有安装 torch 的环境不就没法解析样例配置文件了么，这样的配置文件还叫配置文件么？不要着急，这部分的内容我们会在后面介绍。
```

docs/zh_cn/advanced_tutorials/config.md
```diff
@@ -364,12 +368,447 @@ a=1
...
+
+        # _base_ = [./optimizer.py]
+
+    .. code-tab:: python 纯文本风格继承
+
+        if '_base_':
+            from .optimizer import *
+```
+
+纯 python 风格的配置文件通过 import 语法来实现继承，这样做的好处是，我们可以直接跳转到被继承的配置文件中，方便阅读和跳转。变量的继承规则（增删改查）和 完全对齐 Python 语法，例如我想修改 base 配置文件中 optimizer 的学习率:
```
zhouzaida, line 473:
```suggestion
纯 python 风格的配置文件通过 import 语法来实现继承，这样做的好处是，我们可以直接跳转到被继承的配置文件中，方便阅读和跳转。变量的继承规则（增删改查）完全对齐 Python 语法，例如我想修改 base 配置文件中 optimizer 的学习率：
```

docs/zh_cn/advanced_tutorials/config.md
```diff
@@ -364,12 +368,447 @@ a=1
...
+
+可以看到，纯 python 风格导出的 type 字段会包含模块的全量信息。导出的配置文件也可以被直接加载，通过注册器来构建实例。
+
+### 什么是 lazy import
+
+看到这你可能会吐槽，这纯 python 风格的配置文件感觉就像是用纯 python 语法来组织配置文件嘛。这样我哪还需要配置类，直接用 python 语法来导入配置文件不就好了。如果你有这样的感受，那真是一件值得庆祝的事，因为这正是我们想要的效果。
+
+正如前面所提到的，解析配置文件需要依赖配置文件中引用的三方库，这其实是一件非常不合理的事。例如我基于 MMagic 训练了一个模型，想使用 MMDeploy 的 onnxruntime 后端部署。由于部署环境中没有 torch，而配置文件解析过程中需要 torch，这就导致了我无法直接使用 MMagic 的配置文件作为部署的配置，这是非常不方便的。为了解决这个问题，我们引入了 lazy_import 的概念。
+
+要聊 lazy_import 的具体实现是一件比较复杂的事，在此我们仅对其功能做简要介绍。lazy_import 的核心思想是，将配置文件中的 import 语句延迟到配置文件被解析时才执行，这样就可以避免配置文件中的 import 语句导致的三方库依赖问题。配置文件解析过程时，python 解释器实际执行的等效代码如下
```
zhouzaida, line 573:
```suggestion
要聊 lazy_import 的具体实现是一件比较复杂的事，在此我们仅对其功能做简要介绍。lazy_import 的核心思想是，将配置文件中的 import 语句延迟到配置文件被解析时才执行，这样就可以避免配置文件中的 import 语句导致的三方库依赖问题。配置文件解析过程时，Python 解释器实际执行的等效代码如下
```

zhouzaida: Either unify to Python or python.

docs/zh_cn/advanced_tutorials/config.md
```diff
@@ -364,12 +368,447 @@ a=1
...
+        optimizer = dict(type=SGD)
+
+    .. code-tab:: python 通过配置类，python 解释器实际执行的代码
+
+        lazy_obj = LazyObject('torch.optim', 'SGD')
+
+        optimizer = dict(type=lazy_obj)
+```
+
+LazyObject 作为 `Config` 模块的內部类型，无法被用户访直接。用户在访问 type 字段时，会经过一系列的转换，将 `LazyObject` 转化成真正的 `torch.optim.SGD` 类型。这样一来，配置文件的解析不会触发三方库的导入，而用户使用配置文件时，又可以正常访问三方库的类型。
```
zhouzaida, line 591:
```suggestion
LazyObject 作为 `Config` 模块的內部类型，无法被用户直接访问。用户在访问 type 字段时，会经过一系列的转换，将 `LazyObject` 转化成真正的 `torch.optim.SGD` 类型。这样一来，配置文件的解析不会触发三方库的导入，而用户使用配置文件时，又可以正常访问三方库的类型。
```

docs/zh_cn/advanced_tutorials/config.md
```diff
@@ -364,12 +368,447 @@ a=1
...
+
+要想访问 `LazyObject` 的内部类型，可以通过 `Config.to_dict` 接口：
+
+```python
+cfg = Config.fromfile('optimizer.py').to_dict()
+print(type(cfg['optimizer']['type']))
+# mmengine.config.lazy.LazyObject
+```
+
+此时访问得到的 type 就是 LazyObject 类型。
```
zhouzaida, line 601:
```suggestion
此时得到的 type 就是 LazyObject 类型。
```

docs/zh_cn/advanced_tutorials/config.md
```diff
@@ -364,12 +368,447 @@ a=1
...
+
+然而对于 base 文件的继承（导入，import），我们不能够采取 lazy import 的策略，这是因为我们希望解析后的配置文件能够包含 base 配置文件定义的字段，需要真正的触发 import。因此我们对 base 文件的导入加了一层限制，即必须在 `if '_base_'` 的代码块中导入。
+
+### 功能限制
+
+1. 不能在配置文件中定义函数、类等
+2. 配置文件名必须符合 python 模块名的命名规范，即只能包含字母、数字、下划线，且不能以数字开头
+3. 导入 base 配置文件中的变量时候，例如 `from ._base_.alpha import beta`，此处的 `alpha` 必须是模块（module）名，即 python 文件，而能不是含有 `__init__.py` 的包（package）名
```
zhouzaida, line 609:
```suggestion
3. 导入 base 配置文件中的变量时，例如 `from ._base_.alpha import beta`，此处的 `alpha` 必须是模块（module）名，即 python 文件，而能不是含有 `__init__.py` 的包（package）名
```

docs/zh_cn/advanced_tutorials/config.md
```diff
@@ -364,12 +368,447 @@ a=1
...
+### 迁移指南
+
+从纯文本风格的配置文件迁移到纯 python 风格的配置文件，需要遵守以下规则：
+
+1. type 从字符串替换成具体的类：
+
+   - 代码不依赖 type 字段是字符串，且没有对 type 字段做特殊处理，则可以将字符串类型的 type 替换成具体的类，并在配置文件的开头导入该类
+   - 代码依赖 type 字段是字符串，则需要修改代码，或保持原有的字符串格式的 type
+
+2. rename 配置文件，配置文件命名需要符合 python 模块名的命名规范，即只能包含字母、数字、下划线，且不能以数字开头
```
zhouzaida, line 621:
```suggestion
2. 重命名配置文件，配置文件命名需要符合 python 模块名的命名规范，即只能包含字母、数字、下划线，且不能以数字开头
```

tests/data/config/lazy_module_config/_base_/base_model.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from mmengine.testing.runner_test_case import ToyModel
+
+model = dict(type=ToyModel)
```
zhouzaida, line 4:
```suggestion
model = dict(type=ToyModel)

```

tests/data/config/lazy_module_config/error_mix_using1.py
```diff
@@ -0,0 +1,2 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+_base_ = './toy_model.py'
```
zhouzaida, line 2:
```suggestion
_base_ = './toy_model.py'

```

tests/test_config/test_lazy.py
```diff
@@ -0,0 +1,129 @@
...
+import numpy.compat
+import numpy.linalg as linalg
+
+import mmengine
+from mmengine.config.lazy import LazyAttr, LazyObject
+from mmengine.config.utils import ImportTransformer, _gather_abs_import_lazyobj
+from mmengine.fileio import LocalBackend, PetrelBackend
+
+
+class TestTransform(TestCase):
```
zhouzaida, line 18:
```suggestion
class TestImportTransformer(TestCase):
```

docs/en/advanced_tutorials/config.md
```diff
@@ -370,10 +370,441 @@ a=1
 b=2
 ```
 
+## A Pure Python-style Configuration File
+
+In the previous tutorial, we introduced how to use configuration files to build modules with registrars and how to use `_base_` to inherit configuration files. These pure text-style configuration files can satisfy most of our development needs and some module aliases can greatly simplify the configuration files (e.g. `ResNet` can refer to `mmcls.models.ResNet`). However, there are also some disadvantages:
```
zhouzaida, line 375:
```suggestion
In the previous tutorial, we introduced how to use configuration files to build modules with registry and how to use `_base_` to inherit configuration files. These pure text-style configuration files can satisfy most of our development needs and some module aliases can greatly simplify the configuration files (e.g. `ResNet` can refer to `mmcls.models.ResNet`). However, there are also some disadvantages:
```

docs/en/advanced_tutorials/config.md
```diff
@@ -370,10 +370,441 @@ a=1
...
+
+1. In the configuration file, the `type` field is specified by a string, and IDE cannot directly jump to the corresponding class definition, which is not conducive to code reading and jumping.
+2. The inheritance of configuration files is also specified by a string, and IDE cannot directly jump to the inherited file. When the inheritance structure of the configuration file is complex, it is not conducive to reading and jumping of the configuration file.
+3. The inheritance rules are relatively implicit, and beginners find it difficult to understand how the configuration file merges variables with the same fields and derives special syntax such as `_delete_`, resulting in a higher learning cost.
+4. It is easy for users to forget to register the module and cause `module not found` errors.
+5. In the yet-to-be-mentioned cross-database inheritance, the introduction of the scope makes the inheritance rules of the configuration file more complicated, and beginners find it difficult to understand.
```
zhouzaida, line 381:
```suggestion
5. In the yet-to-be-mentioned cross-codebase inheritance, the introduction of the scope makes the inheritance rules of the configuration file more complicated, and beginners find it difficult to understand.
```

docs/en/advanced_tutorials/config.md
```diff
@@ -370,10 +370,442 @@ a=1
 b=2
 ```
 
+## A Pure Python style Configuration File
```
ly015, line 373: What do you think about marking this part as a 'Beta' feature?

ly015: And we can add a table of content at the beginning.

ly015:
LGTM

tests/data/config/lazy_module_config/_base_/default_runtime.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+# defaults to use registries in mmcls
```
zhouzaida, line 2: What is the meaning of the comment?

mmengine/registry/registry.py
```diff
@@ -431,8 +431,17 @@ def get(self, key: str) -> Optional[Type]:
...
+                'The key argument of `Registry.get` must be a str, '
+                f'got {type(key)}')
         from ..logging import print_log
-
+        try:
+            obj_cls = locate(key)
+        except Exception:
+            raise RuntimeError(f'Failed to get {key}')
+        if obj_cls is not None:
+            return obj_cls
```
zhouzaida, line 444: Please update the docstring of `get` method.

mmengine/runner/runner.py
```diff
@@ -1384,7 +1386,14 @@ def build_dataloader(dataloader: Union[DataLoader, Dict],
...
             worker_init_fn_type = worker_init_fn_cfg.pop('type')
-            worker_init_fn = FUNCTIONS.get(worker_init_fn_type)
+            if isinstance(worker_init_fn_type, str):
+                worker_init_fn = FUNCTIONS.get(worker_init_fn_type)
+            elif callable(worker_init_fn_type):
+                worker_init_fn = worker_init_fn_type
+            else:
+                raise TypeError(
+                    'type of worker_init_fn should be string or callable '
+                    'object!')
```
zhouzaida, line 1396:
```suggestion
                    'type of worker_init_fn should be string or callable '
                    f'object, but got {type(worker_init_fn)}'
```

mmengine/runner/runner.py
```diff
@@ -367,8 +367,9 @@ def __init__(
         mmengine.mkdir_or_exist(self._log_dir)
         # Used to reset registries location. See :meth:`Registry.build` for
         # more details.
-        self.default_scope = DefaultScope.get_instance(
-            self._experiment_name, scope_name=default_scope)
+        if default_scope is not None:
+            self.default_scope = DefaultScope.get_instance(
+                self._experiment_name, scope_name=default_scope)
```
zhouzaida, line 372: If `default_scope` is None, the attribute `default_scope` will not be set. If users access this attribute, an exception will be thrown.

mmengine/registry/registry.py
HAOCHENYE: Actually, it's strange to implement this code snipte here, and let me do some explanation. If we want to build the model using a configuration like `dict(type='mmengine.model.BaseModel')`, which can be dumped by lazy import config, we need the this code snippet for `Registry.get` to work. In this scenario, the instance obtained from `Registry.get` can be any object, not necessarily a strict class or callable object.

In your case, if `mmengine.model.BaseModel.training` is passed in, a `None` object will be return since `training` is not a class attribute for `BaseModel`

I'll add some docstring and comment here.

mmengine/registry/registry.py
```diff
@@ -431,8 +435,25 @@ def get(self, key: str) -> Optional[Type]:
...
+            raise TypeError(
+                'The key argument of `Registry.get` must be a str, '
+                f'got {type(key)}')
         from ..logging import print_log
 
+        # Actually, it's strange to implement this `try ... except` to get the
+        # object by its name in `Registry.get`. However, If we want to build
+        # the model using a configuration like
+        # `dict(type='mmengine.model.BaseModel')`, which can
+        # be dumped by lazy import config, we need the this code snippet
```
zhouzaida, line 448:
```suggestion
        # be dumped by lazy import config, we need this code snippet
```

mmengine/registry/registry.py
```diff
@@ -387,6 +387,10 @@ def get(self, key: str) -> Optional[Type]:
         The method will first parse :attr:`key` and check whether it contains
         a scope name. The logic to search for :attr:`key`:
 
+        - ``key`` represents the whole object name with its module
```
zhouzaida, line 390: If `key`` represents the whole object name with its module
        information, for example, `mmengine.model.BaseModel`, ``get``
        will directly return the class object :class:`BaseModel`.

        Otherwise, it will first parse :attr:`key` and check whether it contains
        a scope name. The logic to search for :attr:`key`:
        - ``key`` does not contain a scope name, i.e., it is purely a module
          name like "ResNet": :meth:`get` will search for ``ResNet`` from the
          current registry to its parent or ancestors until finding it.
        - ``key`` contains a scope name and it is equal to the scope of the
          current registry (e.g., "mmcls"), e.g., "mmcls.ResNet": :meth:`get`
          will only search for ``ResNet`` in the current registry.
        - ``key`` contains a scope name and it is not equal to the scope of
          the current registry (e.g., "mmdet"), e.g., "mmcls.FCNet": If the
          scope exists in its children, :meth:`get` will get "FCNet" from
          them. If not, :meth:`get` will first get the root registry and root
          registry call its own :meth:`get` method.

docs/en/advanced_tutorials/config.md
```diff
@@ -25,6 +50,10 @@ wget https://raw.githubusercontent.com/open-mmlab/mmengine/main/docs/resources/c
 wget https://raw.githubusercontent.com/open-mmlab/mmengine/main/docs/resources/config/modify_base_var.py
 ```
 
+```{note}
+The `Config` supports two styles of configuration files: text style and pure Python style. Each has its own characteristics while maintaining a unified interface for calling. For users who are not familiar with the basic usage of the `Config`, it is recommended to start reading from the section on [Read the configuration file](#read-the-configuration-file) to understand the functionality of the `Config` and the syntax of text style configuration files. In some cases, the syntax of text style configuration files is more concise and compatible with different formats such as `json` and `yaml`. If you prefer a more flexible syntax for configuration files, it is recommended to use the [Pure Python Style Configuration Files" (beta)](#a-pure-python-style-configuration-file-beta).
```
zhouzaida, line 54:
```suggestion
The `Config` supports two styles of configuration files: text style and pure Python style. Each has its own characteristics while maintaining a unified interface for calling. For users who are not familiar with the basic usage of the `Config`, it is recommended to start reading from the section on [Read the configuration file](#read-the-configuration-file) to understand the functionality of the `Config` and the syntax of text style configuration files. In some cases, the syntax of text style configuration files is more concise and compatible with different formats such as `json` and `yaml`. If you prefer a more flexible syntax for configuration files, it is recommended to use the [Pure Python Style Configuration Files (beta)](#a-pure-python-style-configuration-file-beta).
```

zhouzaida (approved):

PR #1072, opened 2023-04-12
[Feature] add auto_step flag in WandbVisBackend

i-aki-y:
## Motivation

The wandb backend uses a mechanism of the auto-increment built in the `wandb.log`.
But this step information will be reset to zero after the resumption.
So when I resume the training, I get the following discontinuous graph.

resume_auto_step

This behavior is a little confusing, I think the graphs should be continuous before and after the resumption.

resume_manual_step

## Modification

To fix this, I added an `auto_step` flag that allows the user to select the increment method.
- auto_step == True: Keep the current behavior (Default)
- auto_step == False: Use the step given by an argument. Note that other loggers also use this argument as the step.

## BC-breaking (Optional)

No. Default auto_step is set to True which is consistent with the current behavior.

## Use cases (Optional)

This makes wandb graph natural before and after the resumption (like the second image above).

## Checklist

### DIFF
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -371,8 +371,9 @@ class WandbVisBackend(BaseVisBackend):
         auto_step: (bool) Whether use auto step increment built-in
             ``wandb.log`` or not. The built-in step will be reset to zero when
             the training is resumed. Default: True
-            Note that if you set False, you also should set
-            ``log_metric_by_epoch=False`` in the LoggerHook.
+
+    Note:
+        If ``auto_step`` is False, ``log_metric_by_epoch`` in :cls:`mmengine.hook.LoggerHook` needs to be set to ``False`` too.
     """
 
     def __init__(self,
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -372,10 +372,6 @@ class WandbVisBackend(BaseVisBackend):
             ``wandb.log`` or not. The built-in step will be reset to zero when
             the training is resumed. Default: True
             New in version 0.7.3
-
-    Note:
-        If ``auto_step`` is False, ``log_metric_by_epoch`` in
-        :cls:`mmengine.hook.LoggerHook` needs to be set to ``False`` too.
     """
 
     def __init__(self,
@@ -393,6 +389,7 @@ class WandbVisBackend(BaseVisBackend):
         self._log_code_name = log_code_name
         self._watch_kwargs = watch_kwargs if watch_kwargs is not None else {}
         self._auto_step = auto_step
+        self._last_step = 0
 
     def _init_env(self):
         """Setup env for wandb."""
@@ -465,6 +462,9 @@ class WandbVisBackend(BaseVisBackend):
         if self._auto_step:
             self._wandb.log({name: image}, commit=self._commit)
         else:
+            if self._last_step > step:
+                step = self._last_step + 1
+            self._last_step = step
             self._wandb.log({name: image}, commit=self._commit, step=step)
 
     @force_init_env
@@ -484,6 +484,9 @@ class WandbVisBackend(BaseVisBackend):
         if self._auto_step:
             self._wandb.log({name: value}, commit=self._commit)
         else:
+            if self._last_step > step:
+                step = self._last_step + 1
+            self._last_step = step
             self._wandb.log({name: value}, commit=self._commit, step=step)
 
     @force_init_env
@@ -505,6 +508,9 @@ class WandbVisBackend(BaseVisBackend):
         if self._auto_step:
             self._wandb.log(scalar_dict, commit=self._commit)
         else:
+            if self._last_step > step:
+                step = self._last_step + 1
+            self._last_step = step
             self._wandb.log(scalar_dict, commit=self._commit, step=step)
 
     def close(self) -> None:
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -368,8 +368,8 @@ class WandbVisBackend(BaseVisBackend):
             New in version 0.3.0.
         watch_kwargs (optional, dict): Agurments for ``wandb.watch``.
             New in version 0.4.0.
-        auto_step: (bool) Whether use auto step increment builtin
-            wandb.log or not. The built-in step will be reset to zero when
+        auto_step: (bool) Whether use auto step increment built-in
+            ``wandb.log`` or not. The built-in step will be reset to zero when
             the training is resumed. Default: True
     """
 
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -371,6 +371,8 @@ class WandbVisBackend(BaseVisBackend):
         auto_step: (bool) Whether use auto step increment built-in
             ``wandb.log`` or not. The built-in step will be reset to zero when
             the training is resumed. Default: True
+            Note that if you set False, you also should set
+            ``log_metric_by_epoch=False`` in the LoggerHook.
     """
 
     def __init__(self,
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -374,7 +374,8 @@ class WandbVisBackend(BaseVisBackend):
             New in version 0.7.3
 
     Note:
-        If ``auto_step`` is False, ``log_metric_by_epoch`` in :cls:`mmengine.hook.LoggerHook` needs to be set to ``False`` too.
+        If ``auto_step`` is False, ``log_metric_by_epoch`` in
+        :cls:`mmengine.hook.LoggerHook` needs to be set to ``False`` too.
     """
 
     def __init__(self,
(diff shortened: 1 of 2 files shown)


mmengine/visualization/vis_backend.py
```diff
@@ -371,6 +371,8 @@ class WandbVisBackend(BaseVisBackend):
         auto_step: (bool) Whether use auto step increment built-in
             ``wandb.log`` or not. The built-in step will be reset to zero when
             the training is resumed. Default: True
+            Note that if you set False, you also should set
+            ``log_metric_by_epoch=False`` in the LoggerHook.
```
HAOCHENYE, line 375:
```suggestion

    Note:
        If ``auto_step`` is False, ``log_metric_by_epoch`` in :cls:`mmengine.hook.LoggerHook` needs to be set to ``False`` too.
```

mmengine/visualization/vis_backend.py
```diff
@@ -368,6 +368,11 @@ class WandbVisBackend(BaseVisBackend):
             New in version 0.3.0.
         watch_kwargs (optional, dict): Agurments for ``wandb.watch``.
             New in version 0.4.0.
+        auto_step: (bool) Whether use auto step increment built-in
+            ``wandb.log`` or not. The built-in step will be reset to zero when
+            the training is resumed. Default: True
```
HAOCHENYE, line 373:
```suggestion
        auto_step: (bool) Whether use auto step increment built-in
            ``wandb.log`` or not. The built-in step will be reset to zero when
            the training is resumed. Default: True
            New in version 0.7.3
```

HAOCHENYE (commented):
Thanks for your contribution. The overall modification LGTM :smile: .

i-aki-y:
@HAOCHENYE  Thank you for your review.

mmengine/visualization/vis_backend.py
```diff
@@ -368,6 +368,13 @@ class WandbVisBackend(BaseVisBackend):
...
+            New in version 0.7.3
+
+    Note:
+        If ``auto_step`` is False, ``log_metric_by_epoch`` in :cls:`mmengine.hook.LoggerHook` needs to be set to ``False`` too.
```
HAOCHENYE, line 377: Needs a line break here.

mmengine/visualization/vis_backend.py
```diff
@@ -465,10 +478,13 @@ def add_scalar(self,
...
-            step (int): Useless parameter. Wandb does not
-                need this parameter. Defaults to 0.
+            step (int): Global step value to record. Defaults to 0.
+                This will be ignored if auto_step is True.
         """
-        self._wandb.log({name: value}, commit=self._commit)
+        if self._auto_step:
+            self._wandb.log({name: value}, commit=self._commit)
+        else:
+            self._wandb.log({name: value}, commit=self._commit, step=step)
```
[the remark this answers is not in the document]
i-aki-y: @zhouzaida  thank you for your review.

To make it work, we need to change the default value of the `log_metric_by_epoch` in the `LoggerHook` as described in the "Note" section in this docstring.
I am not sure I can change the default value like `log_metric_by_epoch` in this PR.

The followings are the reason why `log_metric_by_epoch=False` is needed.

First, the `wandb.log` only supports monotonically increasing step sequences.
(See, https://docs.wandb.ai/guides/track/log/logging-faqs#why-am-i-seeing-fewer-data-points-than-i-logged )
If we set an earlier step in `wandb.log`, the `wandb` looks to ignore the log.
In the `LoggerHook`, the `runner.iter` is used as the step, but there is an exception, the `after_val_epoch` uses `runner.epoch` by default. 
Since almost always the `runner.epoch < runner.iter` is satisfied, we need to set `log_metric_by_epoch=False` to avoid the `wandb` ignoring the validation metrics.


One idea to remove the `auto_step` flag is to add a new attribute, `last_step`, in the `WandbVisBackend`. 
When the step < self.last_step is satisfied (this is probably runner.epoch), replace it with the `last_step`.

```py
if step < self.last_step:
    step = self.last_step
else:
    self.last_step = step
self._wandb.log(..., step=step)
```

I think this can ensure the step is always monotonically increasing. This may be more robust and better than `auto_flag` approach because we do not need to care about the value of `log_metric_by_epoch`.
Should I try this version?

i-aki-y:
Sorry, I am reconsidering this PR because my first version is not user-friendly a little. In my first version, the `auto_step` interacts with `LoggerHook`'s `log_metric_by_epoch`, but I think this interaction is non-trivial for many users. And if users misconfigure them, the `wandb` will ignore the validation metrics without any errors. This misconfiguration could often happen in a complicated system like mmdet and confuse users.

mmengine/visualization/vis_backend.py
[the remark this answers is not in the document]
zhouzaida: Hi @i-aki-y , thanks for your very clear explanation and I got your point. Due to the feature of wandb.log, we should increase the `step` monotonically. The idea you proposed looks good to me.

zhouzaida:
> Sorry, I am reconsidering this PR because my first version is not user-friendly. In my first version, the `auto_step` interacts with `LoggerHook`'s `log_metric_by_epoch`, but I think this interaction is non-trivial for many users. And if users misconfigure them, the `wandb` will ignore the validation metrics without any errors. This misconfiguration could often happen in a complicated system like mmdet and confuse users.

Hi @i-aki-y , the idea your proposed above looks good to me.

i-aki-y:
@zhouzaida @HAOCHENYE I revised the PR, I would appreciate it if you review it.

As I answered @zhouzaida's comment, I introduced the `self._last_step` (`+ 1` is needed) to avoid an invalid step being passed to `wandb.log`. As a result, the implementation becomes a bit complicated, but I think this is worth doing. Now I could remove the note for the `log_metric_by_epoch`.

I still left the `auto_step` flag to keep the previous behavior. But switching the `auto_flag` changes only the unit of step (from the "count of `wandb.log` calls" to the `runner.iter`) and resumption behavior.
If such a change in the step's unit is acceptable, let me know, and I will remove the `auto_flag` to always use `wandb.log(...,step=step)`.

HAOCHENYE:
Hi! The modification makes sense to me. Refers to https://github.com/wandb/wandb/issues/2642 and https://docs.wandb.ai/ref/python/run#define_metric, I found that WandB can support logging data on different x axis like that:

```python
import wandb

# 1. Start a W&B run
wandb.init(project='test')

# 2. Save model inputs and hyperparameters
config = wandb.config
config.learning_rate = 0.01
wandb.define_metric('loss1', step_metric='iter', step_sync=False)
wandb.define_metric('loss2', step_metric='epoch', step_sync=False)
# Model training code here ...

# 3. Log metrics over time to visualize performance
for i in range(100):
    wandb.log({"loss1": i, 'iter': i})
    if i % 10 == 0:
        wandb.log({"loss2": i, 'epoch': i // 10})
        wandb.log({"loss3": i * 10, 'epoch': i // 10})
```
Will this make it possible to record epoch-based data?

i-aki-y:
@HAOCHENYE  Thank you for pointing interesting feature.

It looks to work as expected!

Unfortunately, the current `define_metric_cfg` option of the `WandbVisBackend` only supports a case of `define_metric(metric, summary=summary)`, but if we can fix this and makes include the "iter" (and “epoch" when by_epoch==True), to the `tag` of the log output, we can control the behavior of the wandb plot more flexibly.

HAOCHENYE:
> Author

I think this kind of modification `define_metric_cfg={'coco/bbox_mAP': 'max', step_type: 'epoch'}` is OK

i-aki-y:
@HAOCHENYE  thank you for the advice.

I think we should allow setting multiple define_metrics in the cfg.
Can I re-define the existing `define_metric_cfg` as a list of dicts from a dict, or should I introduce a new argument like `define_metric_cfg"s”`?

I think the following usage:

```
vis_backends = [
    dict(type='LocalVisBackend'),
    dict(
        type='WandbVisBackend',        
        init_kwargs={...},
        define_metric_cfgs=[
            dict(name="lr", step_metric='iter’),
            ...
            dict(name="coco/bbox_mAP", step_metric='epoch'),
            dict(name="coco/bbox_mAP_50", step_metric='epoch'),            
        ],
    )
]
```

i-aki-y:
@HAOCHENYE @zhouzaida  Since the solution that uses define_metric is based on a different idea and the implementation is also different,  I made a separate PR (#1099).

If the new solution is better I will close this issue.

zhouzaida:
> @HAOCHENYE thank you for the advice.
> 
> I think we should allow setting multiple define_metrics in the cfg. Can I re-define the existing `define_metric_cfg` as a list of dicts from a dict, or should I introduce a new argument like `define_metric_cfg"s”`?
> 
> I think the following usage:
> 
> ```
> vis_backends = [
>     dict(type='LocalVisBackend'),
>     dict(
>         type='WandbVisBackend',        
>         init_kwargs={...},
>         define_metric_cfgs=[
>             dict(name="lr", step_metric='iter’),
>             ...
>             dict(name="coco/bbox_mAP", step_metric='epoch'),
>             dict(name="coco/bbox_mAP_50", step_metric='epoch'),            
>         ],
>     )
> ]
> ```

Hi, we can re-define the existing `define_metric_cfg` which can be a dict or list of dict.

zhouzaida:
> @HAOCHENYE @zhouzaida Since the solution that uses define_metric is based on a different idea and the implementation is also different, I made a separate PR (#1099).
> 
> If the new solution is better I will close this issue.

The solution in #1099 maybe a more elegant solution so we can close this PR. Thanks for your contribution very much.

i-aki-y:
I think PR (https://github.com/open-mmlab/mmengine/pull/1099) is a better solution, so I am closing this PR

PR #972, opened 2023-02-28
[Enhance] Support writing data to `vis_backend` with prefix

HAOCHENYE:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

1. Support log information to various backends(Tensorboard, WandB) with **hierarchy**
2. Count averaged `time` and `data_time` **by epoch during validation**

Enable `log_with_hierarchy` in `LogProcessor`

```python
log_processor = dict(
    log_with_hierarchy=False
)
```

**original layout**:
image

**hierarchy layout**

image

As you can see, logs are divided into `train` and `val`, and the averaged `time`, `data_time` during validation is counted.
## Modification

Please briefly describe what modification is made in this PR.

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repos?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
3. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
4. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
5. The documentation has been modified accordingly, like docstring or example tutorials.

PR #1078, opened 2023-04-16
[Fix] Fix the resuming error caused by HistoryBuffer

HAOCHENYE:

## Motivation

After PR #972, `HistoryBuffer.statistics` will be called with the [custom_keys_copy](https://github.com/open-mmlab/mmengine/pull/972/files#diff-75746b50b571a33d63b68111136d0aa29ac6ac5d788aff12b02a7f329f640728R286). However, the class attributed [`HistoryBuffer._statistics_methods`](https://github.com/HAOCHENYE/mmengine/blob/d132e41a974f6f6e99ca6bdcb9eb56b7d9d2d653/mmengine/logging/history_buffer.py#L32) will not be dumped during saving the checkpoint.  It means the resumed `HistoryBuffer`s can not call the `statistic` methods since they have an empty `_statistic_methods`

This error can be trigger follow these steps:

1. set `log_with_hierarchy=True` for the `log_processor` and start training
2. Stop the training until the next checkpoint has been saved after the prior validation. At this time, All `HistoryBuffer`s including `val/time` and `val/data_time` will be dumped without `_statistics_methods`.
3. Resume the training. The `HistoryBuffer`s will be loaded without `_statistics_methods`.
4. During validation, the `HistoryBuffer` of `val/data_time` and `val/time` will call `statistics` method without `_statistics_methods`, and an error will be raised.

## Modification

1. Make `HistoryBuffer` Dump the `_statistics_methods`.
2. Refine the logic of `log_processor`

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

4. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
5. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
6. The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/examples/distributed_training.py
+++ b/examples/distributed_training.py
@@ -92,7 +92,6 @@ def main():
         val_dataloader=val_dataloader,
         val_cfg=dict(),
         val_evaluator=dict(type=Accuracy),
-        resume=True,
         launcher=args.launcher,
     )
     runner.train()
--- a/examples/distributed_training.py
+++ b/examples/distributed_training.py
@@ -92,6 +92,7 @@ def main():
         val_dataloader=val_dataloader,
         val_cfg=dict(),
         val_evaluator=dict(type=Accuracy),
+        resume=True,
         launcher=args.launcher,
     )
     runner.train()
--- a/tests/test_runner/test_runner.py
+++ b/tests/test_runner/test_runner.py
@@ -21,7 +21,7 @@ from mmengine.evaluator import BaseMetric, Evaluator
 from mmengine.hooks import (CheckpointHook, DistSamplerSeedHook, Hook,
                             IterTimerHook, LoggerHook, ParamSchedulerHook,
                             RuntimeInfoHook)
-from mmengine.logging import MessageHub, MMLogger
+from mmengine.logging import HistoryBuffer, MessageHub, MMLogger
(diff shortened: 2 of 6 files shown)


examples/distributed_training.py
```diff
@@ -92,6 +92,7 @@ def main():
         val_dataloader=val_dataloader,
         val_cfg=dict(),
         val_evaluator=dict(type=Accuracy),
+        resume=True,
```
zhouzaida, line 95:
```suggestion
```

tests/test_runner/test_runner.py
```diff
@@ -2304,6 +2304,10 @@ def test_checkpoint(self):
         message_hub.load_state_dict(ckpt['message_hub'])
         self.assertEqual(message_hub.get_info('epoch'), 0)
         self.assertEqual(message_hub.get_info('iter'), 11)
+        # 2.1.2 check class attribute _statistic_methods can be saved
+        HistoryBuffer._statistics_methods.clear()
+        ckpt = torch.load(path)
+        self.assertIn('min', HistoryBuffer._statistics_methods)
```
zhouzaida, line 2310: It seems that the ckpt has not used.

zhouzaida (approved):

PR #1099, opened 2023-04-23
[Feature] Add support for full wandb's define_metric arguments

i-aki-y:
## Motivation

In this PR, add support for other original arguments of wandb.define_metric described in: https://docs.wandb.ai/ref/python/run#define_metric
(The current `WandbVisBackend` implementation supports only the case of the `summary` metric.)

Especially the `step_metric` is helpful for defining the x-axis more flexibly.
Specifying the "iter" or "epoch" as an x-axis, we can solve the issue of resumption discussed in issue #1072. I think this PR is a better solution because this does not need additional variables like `auto_step` nor`last_step` nor extra logic and uses only the wandb's feature.

Furthermore, users can specify different x-axes depending on the metrics in the same experiment. See below:

iter_epoch

## Modification

To allow the visualizers to use the `iter` as a step, I added an `iter` key in the tag of the log_processor. 
And I added a new argument (`define_metric_cfgs`) in the WandbVisBackend for users to configure the metric.

Since all visualizers use the same tag, this affects other non-wandb visualizers' behavior.
I think this modification would not cause any negative effect on other visualizers. 
But if it causes something bad, I will add a flag like 'add_iter' in the  log_processor for user control to include `iter` or not.

## BC-breaking (Optional)

As I described above, this PR added a new key, "iter," in the output of the log_processor.
A new plot for the "iter" might appear in some visualization, but  I think it does not cause serious problems.

## Use cases (Optional)

Expected usage is defining all the metrics used by the experiment:

```
vis_backends = [
    dict(type='LocalVisBackend'),
    dict(
        type='WandbVisBackend',        
        init_kwargs={...},
        define_metric_cfgs=[
            dict(name="lr", step_metric='iter’),
            ...
            dict(name="coco/bbox_mAP", step_metric='epoch'),
            dict(name="coco/bbox_mAP_50", step_metric='epoch'),            
        ],
    )
]
```

Providing a standard preset of the define_metric_cfgs as default (ex., `configs/_base_/wandb_metrics.py`) might be great, but I don't think it is indispensable. It is not difficult, though cumbersome, for users to define metrics individually for a particular project. And defining all possible metrics in advance is a bit difficult because the metrics would change depending on the models and tasks.

If the user missed defining some metrics, wandb would use the default step as the x-axis and loose nothing.
We can fix the X-axis later in the Web UI.

switch_x

## Checklist

### DIFF
--- a/mmengine/runner/log_processor.py
+++ b/mmengine/runner/log_processor.py
@@ -201,6 +201,8 @@ class LogProcessor:
                 cur_iter_str = str(batch_idx + 1).rjust(
                     len(str(dataloader_len)))
                 log_str = (f'Iter({mode}) [{cur_iter_str}/{dataloader_len}]  ')
+        # Add global iter.
+        tag['iter'] = runner.iter + 1
         # Concatenate lr, momentum string with log header.
         log_str += f'{lr_str}  '
         # If IterTimerHook used in runner, eta, time, and data_time should be
--- a/mmengine/runner/log_processor.py
+++ b/mmengine/runner/log_processor.py
@@ -202,7 +202,10 @@ class LogProcessor:
                     len(str(dataloader_len)))
                 log_str = (f'Iter({mode}) [{cur_iter_str}/{dataloader_len}]  ')
         # Add global iter.
-        tag['iter'] = runner.iter + 1
+        if isinstance(runner._train_loop, dict) or runner._train_loop is None:
+            tag['iter'] = 0
+        else:
+            tag['iter'] = runner.iter + 1
         # Concatenate lr, momentum string with log header.
         log_str += f'{lr_str}  '
         # If IterTimerHook used in runner, eta, time, and data_time should be
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -347,15 +347,15 @@ class WandbVisBackend(BaseVisBackend):
             input parameters.
             See `wandb.init <https://docs.wandb.ai/ref/python/init>`_ for
             details. Defaults to None.
-        define_metric_cfg (dict | list[dict], optional):
+        define_metric_cfg (dict or list[dict], optional):
             When a dict is set, it is a dict of metrics and summary for
-            wandb.define_metric.
+            ``wandb.define_metric``.
             The key is metric and the value is summary.
             When a list is set, each dict should be a valid argument of
             the ``define_metric``.
-            When ``define_metric_cfg={'coco/bbox_mAP': 'max'}``,
-            The maximum value of ``coco/bbox_mAP`` is logged on wandb UI.
-            When ``define_metric_cfg=[dict(name="loss",
+            For example, ``define_metric_cfg={'coco/bbox_mAP': 'max'}``,
+            means the maximum value of ``coco/bbox_mAP`` is logged on wandb UI.
+            When ``define_metric_cfg=[dict(name='loss',
             step_metric='epoch')]``,
             the "loss” will be plotted against the epoch.
             See `wandb define_metric <https://docs.wandb.ai/ref/python/
@@ -379,7 +379,7 @@ class WandbVisBackend(BaseVisBackend):
     def __init__(self,
                  save_dir: str,
                  init_kwargs: Optional[dict] = None,
-                 define_metric_cfg: Optional[Union[dict, list]] = None,
+                 define_metric_cfg: Union[dict, list, None] = None,
                  commit: Optional[bool] = True,
                  log_code_name: Optional[str] = None,
                  watch_kwargs: Optional[dict] = None):
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -383,7 +383,7 @@ class WandbVisBackend(BaseVisBackend):
                  save_dir: str,
                  init_kwargs: Optional[dict] = None,
                  define_metric_cfg: Optional[dict] = None,
-                 define_metric_cfgs: Optional[dict] = None,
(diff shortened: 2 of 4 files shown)


i-aki-y:
Updated: allow define_metric_cfg to accept dict and list of dict, and remove added define_metric_cfgs argument.

I noticed that the define_metric support glob so that we can use the following setting:

```
 define_metric_cfgs=[
            dict(name="loss*", step_metric='iter’),
            dict(name="coco/*", step_metric='epoch'),
            ...
        ],
```

https://docs.wandb.ai/guides/track/log/customize-logging-axes

mmengine/visualization/vis_backend.py
```diff
@@ -347,11 +347,17 @@ class WandbVisBackend(BaseVisBackend):
...
+        define_metric_cfg (dict | list[dict], optional):
+            When a dict is set, it is a dict of metrics and summary for
+            wandb.define_metric.
             The key is metric and the value is summary.
+            When a list is set, each dict should be a valid argument of
+            the ``define_metric``.
             When ``define_metric_cfg={'coco/bbox_mAP': 'max'}``,
             The maximum value of ``coco/bbox_mAP`` is logged on wandb UI.
+            When ``define_metric_cfg=[dict(name="loss",
+            step_metric='epoch')]``,
```
HAOCHENYE, line 359:
```suggestion
            When ``define_metric_cfg=[dict(name='loss',
            step_metric='epoch')]``,
```

mmengine/visualization/vis_backend.py
```diff
@@ -373,7 +379,7 @@ class WandbVisBackend(BaseVisBackend):
     def __init__(self,
                  save_dir: str,
                  init_kwargs: Optional[dict] = None,
-                 define_metric_cfg: Optional[dict] = None,
+                 define_metric_cfg: Optional[Union[dict, list]] = None,
```
HAOCHENYE, line 382:
```suggestion
                 define_metric_cfg: Union[dict, list, None] = None,
```

HAOCHENYE (commented):
Thanks for your contribution! But should we make `self.wandb.log` accept the `step` argument in this PR? Will the current modification resolve the resume problem?

i-aki-y:
@HAOCHENYE  Thank you for your review.

>Thanks for your contribution! But should we make self.wandb.log accept the step argument in this PR? Will the current modification resolve the resume problem?

We can fix the continuity of the resumed plot by changing the X-axis from the 'step' to 'iter' (or epoch).

fix_resume

In this PR, we include the 'iter' in the log target so that the wandb knows how to map from 'step' to 'iter' even if the user did not set define_metric in advance.

This means that the user needs to change X-axis by themselves by using define_metric or Web UI.
For example, the following settings are working for me so far:
```
    define_metric_cfg=[
        dict(name="coco/*", step_metric="epoch"),
        dict(name="*", step_metric="iter"),
    ],
```

HAOCHENYE:
> @HAOCHENYE Thank you for your review.
> 
> > Thanks for your contribution! But should we make self.wandb.log accept the step argument in this PR? Will the current modification resolve the resume problem?
> 
> We can fix the continuity of the resumed plot by changing the X-axis from the 'step' to 'iter' (or epoch).
> 
> fix_resume
> 
> In this PR, we include the 'iter' in the log target so that the wandb knows how to map from 'step' to 'iter' even if the user did not set define_metric in advance.
> 
> This means that the user needs to change X-axis by themselves by using define_metric or Web UI. For example, the following settings are working for me so far:
> 
> ```
>     define_metric_cfg=[
>         dict(name="coco/*", step_metric="epoch"),
>         dict(name="*", step_metric="iter"),
>     ],
> ```

Thanks for your detailed explanation :smile: ! I've got it!

mmengine/runner/log_processor.py
```diff
@@ -201,6 +201,8 @@ def get_log_after_iter(self, runner, batch_idx: int,
                 cur_iter_str = str(batch_idx + 1).rjust(
                     len(str(dataloader_len)))
                 log_str = (f'Iter({mode}) [{cur_iter_str}/{dataloader_len}]  ')
+        # Add global iter.
+        tag['iter'] = runner.iter + 1
```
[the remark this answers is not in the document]
HAOCHENYE: Suggest assigning `iter` like this:

```python
if isinstance(runner._train_loop, dict) or runner._train_loop is None:
     tag['iter'] = 0
else:
    tag['iter'] = runner.iter + 1
```

i-aki-y:
@HAOCHENYE  Thank you for your comment.
I added a code to avoid log_processor to build train_loop.

> Suggest assigning iter like this:
> if isinstance(runner._train_loop, dict) or runner._train_loop is None:
>      tag['iter'] = 0
> else:
>     tag['iter'] = runner.iter + 1

mmengine/visualization/vis_backend.py
```diff
@@ -347,11 +347,17 @@ class WandbVisBackend(BaseVisBackend):
...
             The key is metric and the value is summary.
-            When ``define_metric_cfg={'coco/bbox_mAP': 'max'}``,
-            The maximum value of ``coco/bbox_mAP`` is logged on wandb UI.
+            When a list is set, each dict should be a valid argument of
+            the ``define_metric``.
+            For example, ``define_metric_cfg={'coco/bbox_mAP': 'max'}``,
+            means the maximum value of ``coco/bbox_mAP`` is logged on wandb UI.
+            When ``define_metric_cfg=[dict(name='loss',
+            step_metric='epoch')]``,
+            the "loss” will be plotted against the epoch.
```
zhouzaida, line 360:
```suggestion
            the "loss" will be plotted against the epoch.
```

HAOCHENYE (approved):

Issue #1111, opened 2023-04-26
[Bug] Register class from child registry into parent


mmeendez8:
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmengine/issues) and [Discussions](https://github.com/open-mmlab/mmengine/discussions) but cannot get the expected help.
- [X] The bug has not been fixed in the latest version(https://github.com/open-mmlab/mmengine).

### Environment

[...]

PR #1114, opened 2023-04-26
Add short explanation about registry scope

mmeendez8:
## Motivation

Try to clarify a bit more how registry works.
https://github.com/open-mmlab/mmengine/issues/1111

## Modification

Added docs

### DIFF
--- a/docs/en/advanced_tutorials/registry.md
+++ b/docs/en/advanced_tutorials/registry.md
@@ -274,13 +274,15 @@ print(output)
 
 ### How does the parent node know about child registry?
 
-When working in our `MMAlpha` it might be necessary to use the `Runner` class defined in MMENGINE. This class is in charge of building most of the objects. If this objects are added to the child registry (`MMAlpha`), how is  `MMEngine` able to find them? It cannot, `MMEngine` needs to switch to the Registry from `MMEngine` to `MMAlpha` according to the scope which is defined in default_runtime.py for searching the target class.
+When working in our `MMAlpha` it might be necessary to use the `Runner` class defined in MMENGINE. This class is in charge of building most of the objects. If these objects are added to the child registry (`MMAlpha`), how is  `MMEngine` able to find them? It cannot, `MMEngine` needs to switch to the Registry from `MMEngine` to `MMAlpha` according to the scope which is defined in default_runtime.py for searching the target class.
 
 We can also init the scope accordingly, see example below:
 
 ```python
 from mmalpha.registry import MODELS
+from mmengine.registry import MODELS as MMENGINE_MODELS
 from mmengine.registry import init_default_scope
+import torch.nn as nn
 
 @MODELS.register_module()
 class LogSoftmax(nn.Module):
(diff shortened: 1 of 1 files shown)


docs/en/advanced_tutorials/registry.md
```diff
@@ -272,6 +272,31 @@ output = model(input)
 print(output)
 ```
 
+### How does the parent node know about child registry?
+
+When working in our `MMAlpha` it might be necessary to use the `Runner` class defined in MMENGINE. This class is in charge of building most of the objects. If this objects are added to the child registry (`MMAlpha`), how is  `MMEngine` able to find them? It cannot, `MMEngine` needs to switch to the Registry from `MMEngine` to `MMAlpha` according to the scope which is defined in default_runtime.py for searching the target class.
```
HAOCHENYE, line 277:
```suggestion
When working in our `MMAlpha` it might be necessary to use the `Runner` class defined in MMEngine. This class is in charge of building most of the objects. If this objects are added to the child registry (`MMAlpha`), how is  `MMEngine` able to find them? It cannot, `MMEngine` needs to switch to the Registry from `MMEngine` to `MMAlpha` according to the `default_scope` argument when building the runner (commonly defined in `default_runtime.py` of downstream repositories) for searching the target class.
```

mmeendez8:
@HAOCHENYE can't we merge this pr?

docs/en/advanced_tutorials/registry.md
```diff
@@ -272,6 +272,31 @@ output = model(input)
...
+@MODELS.register_module()
+class LogSoftmax(nn.Module):
+    def __init__(self, dim=None):
+        super().__init__()
+
+    def forward(self, x):
+        print('call LogSoftmax.forward')
+        return x
+
+assert "LogSoftmax" in MODELS # This fails since registry is looking only in parent class
```
HAOCHENYE, line 294: Hi, I found that this assertion will not False in v0.8.0. I've also tried an old version such as v0.3.0, this assertion will still succeed.  I think a more proper example here is to build the instance by `MODLES.build(dict(type='ToyModel'))`, and it will fail if `init_default_scope('mmalpha')` have not been called before.

mmeendez8: I tried it too and you are right, that was not working as expected. I updated the example, please have a look and let me know what you think :)

docs/en/advanced_tutorials/registry.md
```diff
@@ -272,6 +272,39 @@ output = model(input)
...
+# Works because we are using mmalpha registry
+MODELS.build(dict(type="LogSoftmax"))
+
+# Fails because mmengine registry does not know about stuff registered in mmalpha
+MMENGINE_MODELS.build(dict(type="LogSoftmax"))
+
+# Works because we are using mmalpha registry
+init_default_scope('mmalpha')
+MMENGINE_MODELS.build(dict(type="LogSoftmax"))
+
```
zhouzaida, line 305:
```suggestion
```

zhouzaida (approved):

Issue #1117, opened 2023-04-28
[Feature] Develop the Training example for Detection, Segmentation, and NLP Tasks using MMEngine and MMCV


HAOCHENYE:
### What is the feature?

Develop the Training example for Detection, Segmentation, and NLP Tasks using MMEngine and MMCV referring to this [training example](https://github.com/open-mmlab/mmengine/blob/main/examples/distributed_training.py). 

Different from the complex training pipeline in large repositories, this training example should follow two guides:

1. **As simple as possible**: This example should tell users how to use MMEngine to do these tasks, and should not include too many domain-specific training tricks. The **simple** principle is reflected in many ways:
  a. Simple algorithm
  b. Simple implementation
  c. Easy to start

2. **minimum dependency**: A part from MMEngine, MMCV, and  MMEval and their corresponding dependencies, the example should not depend on other packages (if it is possible), including open-mmlab repositories

PR #1134, opened 2023-05-06
Code example for Object Detection using Mask RCNN

Om-Doiphode:
Developed training example for object detection #1117

### DIFF
--- a/examples/detection_example.py
+++ b/examples/detection_example.py
@@ -1,0 +1,104 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import argparse
+
+import torchvision
+import torchvision.transforms as transforms
+from torch.optim import SGD
+from torchvision.models.detection import maskrcnn_resnet50_fpn
+
+from mmengine.evaluator import BaseMetric
+from mmengine.model import BaseModel
+from mmengine.runner import Runner
+
+
+class MMMaskRCNN(BaseModel):
+
+    def __init__(self):
+        super().__init__()
+        self.model = maskrcnn_resnet50_fpn(pretrained=True)
+
+    def forward(self, images, targets, mode):
+        if mode == 'loss':
+            loss_dict = self.model(images, targets)
+            return loss_dict
+        elif mode == 'predict':
+            predictions = self.model(images)
+            return predictions, targets
+
+
+class Accuracy(BaseMetric):
+
+    def process(self, data_batch, data_samples):
+        predictions, labels = data_samples
+        correct = (predictions.argmax(dim=1) == labels).sum().item()
+        batch_size = labels.size(0)
+        self.results.append({'batch_size': batch_size, 'correct': correct})
+
+    def compute_metrics(self, results):
+        total_correct = sum(item['correct'] for item in results)
+        total_size = sum(item['batch_size'] for item in results)
+        accuracy = 100 * total_correct / total_size
+        return {'accuracy': accuracy}
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Distributed Training')
+    parser.add_argument(
+        '--launcher',
+        choices=['none', 'pytorch', 'slurm', 'mpi'],
+        default='none',
+        help='job launcher')
+    parser.add_argument('--local_rank', type=int, default=0)
+    parser.add_argument(
+        '--batch_size', type=int, default=32, help='batch size for training')
+
+    args = parser.parse_args()
+    return args
+
+
+def main():
+    args = parse_args()
+    norm_cfg = dict(mean=[0.491, 0.482, 0.447], std=[0.202, 0.199, 0.201])
+    train_set = torchvision.datasets.Kitti(
+        'data/kitti',
+        train=True,
+        download=True,
+        transform=transforms.Compose([
+            transforms.RandomCrop(32, padding=4),
+            transforms.RandomHorizontalFlip(),
+            transforms.ToTensor(),
+            transforms.Normalize(**norm_cfg)
+        ]))
+    valid_set = torchvision.datasets.Kitti(
+        'data/kitti',
+        train=False,
+        download=True,
+        transform=transforms.Compose(
+            [transforms.ToTensor(),
+             transforms.Normalize(**norm_cfg)]))
+    train_dataloader = dict(
+        batch_size=args.batch_size,
+        dataset=train_set,
+        sampler=dict(type='DefaultSampler', shuffle=True),
+        collate_fn=dict(type='default_collate'))
+    val_dataloader = dict(
+        batch_size=args.batch_size,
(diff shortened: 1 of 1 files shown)


Om-Doiphode:
@HAOCHENYE, I have issued a PR, please review. Thanks!

HAOCHENYE (commented):
Hi, thanks for your contribution, can you provide the training logs of this example?

BTW, the KITTI dataset seems a little bit large for downloading. Is there a smaller dataset? Maybe a slice of Coco dataset is enough, for example, the [coco128](https://www.kaggle.com/datasets/ultralytics/coco128), and we can divide it into train set and validation set. Although such a small dataset will trigger the overfitting problem, however, I think this is acceptable for an **example**.

It seems this example still uses the `Accuray· Metric, suggest using a iou metric to evaluate the model. [MMEval](https://github.com/open-mmlab/mmeval/tree/9ce87f0f45732b220ff566e3b0c9fabf3c2baf14) has defined a lot of metrics, and you can try to use them.

HAOCHENYE (commented):
Hi, thanks for your contribution, can you provide the training logs of this example?

BTW, the KITTI dataset seems a little bit large for downloading. Is there a smaller dataset? Maybe a slice of Coco dataset is enough, for example, the [coco128](https://www.kaggle.com/datasets/ultralytics/coco128), and we can divide it into train set and validation set. Although such a small dataset will trigger the overfitting problem, however, I think this is acceptable for an **example**.

It seems this example still uses the `Accuray· Metric, suggest using a iou metric to evaluate the model. [MMEval](https://github.com/open-mmlab/mmeval/tree/9ce87f0f45732b220ff566e3b0c9fabf3c2baf14) has defined a lot of metrics, and you can try to use them.

HAOCHENYE:
The current implementation seems a little bit comlicated, the error indicates that the index given by sample does not match with the implementation here:

https://github.com/open-mmlab/mmengine/pull/1134/files#diff-ee540e9d5ba67d4968f7f2640915c3e1e333de63a16bfd6d1620cabfc60c005bR220

Before continue developing with coco128, we need to figure out is coco128 fit for the example:

1. Can we download coco128 in Python code, and then convert it to the target format to reuse the existing data class like `tochvision.datasets.CocoDetection`?
2. Could we train a simpler algorithm such as RetinaNet? I'm not sure does coco128 contain the sufficient label information for instance segmentation tasks
3. coco128 is only a suggestion, any other smaller dataset is acceptable.

Om-Doiphode:
Hi @HAOCHENYE, the issue still persists.
1. The COCO dataset is not available in torchvision library so we need to externally download the dataset.
2. 
image

The following error occurs on using dict()
image

3. 
image
While on using Dataloader from torchvision the following error occurs:
image

Even on using kitti dataset, the above error occurs, can you please guide me on how to solve this problem?

Om-Doiphode:
Hi, @HAOCHENYE, After implementing a custom collate function for COCO dataset, getting the following error:
image

HAOCHENYE:
It seems that your `collate_fn` returns an improper batch data. You can see this [tutorial](https://pytorch.org/docs/stable/data.html#automatic-batching-default) for more information about `collate_fn`

HAOCHENYE:
Besides, if there is no available URL to download coco128, you can also use this dataset:

https://download.openmmlab.com/mmyolo/data/balloon_dataset.zip

and use `torch.hub.download_url_to_file` to download it in the code. However there is only one class in this dataset, you need to change the model structure.

Om-Doiphode:
Hi, @HAOCHENYE, seems like the issue has been solved with the collate function, but getting error while calculating the Accuracy using IOU score. I have defined my own Accuracy class but getting errors, sometimes the model predict lesser number of bounding boxes as compared to the actual bounding boxes and vice versa. Can you please guide me on what to do in this case?

Om-Doiphode:
Hi @HAOCHENYE, the model is working successfully, albeit the accuracy is low. Please review, thanks!

HAOCHENYE (commented):
Hi, please include `coco.py` and `detection_example.py` in the `detection` directory, and provide a `README.md` to tell users how to train this example

examples/coco_utils.py
HAOCHENYE, line 1: If this `util` module is only used for converting the dataset format, I think we can save the converted data in our oss, and download dataset in the example code,  just like `torchvision.dataset` does. Besides, could the [mini-balloon_dataset](https://github.com/open-mmlab/mmyolo/blob/main/tools/misc/download_dataset.py#L93) be used in this example?

HAOCHENYE:
> Hi @HAOCHENYE, the model is working successfully, albeit the accuracy is low. Please review, thanks!

BTW, the official mask-rcnn is designed for detecting 80 different classes, while it appears that the example provided is focused on single class detection. Could this be a contributing factor to the lower accuracy observed?

Om-Doiphode:
Hi, @HAOCHENYE, I have made the readme, please review. Thanks!

Om-Doiphode:
> > Hi @HAOCHENYE, the model is working successfully, albeit the accuracy is low. Please review, thanks!
> 
> BTW, the official mask-rcnn is designed for detecting 80 different classes, while it appears that the example provided is focused on single class detection. Could this be a contributing factor to the lower accuracy observed?


It seems like the dataset I was using wasn't appropriate for this object detection example because I tried training for more epochs, tweaked the hyper parameters still the accuracy didn't increase. So now I have switched to using COOC128. The accuracy is now much better.

Om-Doiphode:
Hi, @HAOCHENYE, is there something else which needs to be done for this PR?

HAOCHENYE (commented):
The purpose of the examples is to demonstrate how convenient it is to build an algorithm pipeline using MMEngine. Specific algorithm implementations, dataset implementations, and metric implementations should not take up too much space.

For example, we directly use `torchvision`'s models and datasets to construct the training process. We can also use `Metric` defined in `MMEval` or `TorchMetric` to calculate validation accuracy. MMEngine serves as a combination of these modules and should not have too many implementation details in the examples.

The current implementation exposes too many details about the implementations, which is not ideal for an example.

Om-Doiphode:
Hi @HAOCHENYE, As per your suggestions I have made the following changes:
1) Removed coco_utils.py file
2) Eliminated usage of torchvision dataset

Furthermore, I am already utilizing the `COCODetection` metric from `mmeval` to calculate accuracy:

image

Please review. Thanks!

examples/detection/README.md
HAOCHENYE, line 1: Hello! Could you provide me with a step-by-step guide on running this example? I find that the section titled "Prepare your dataset" lacks sufficient detail to guide users to run this example.

Om-Doiphode: I have made the necessary changes. Please review. Thanks!

PR #1140, opened 2023-05-10
[Fix] Fix `ProfileHook` cannot profile ddp-training

HAOCHENYE:

## Motivation

PyTorch profiler support profiling distributed program, however, the current implementation of ProfilerHook's method is decorated by `master_only`, which limits profiler capability.

Besides, If the `scheduler` is configured for the ProfilerHook, an error will be raised like this:

```bash
RuntimeError: stack.size() INTERNAL ASSERT FAILED at "../torch/csrc/autograd/profiler_python.cpp":963, please report a bug to PyTorch. Python replay stack is empty.
```

The main reason is that `self.profile.step` will not be called if scheduler is not None:

https://github.com/open-mmlab/mmengine/blob/8a0fae01f5cfdeeec4bdef4ffd175de1093edd92/mmengine/hooks/profiler_hook.py#L226-L236

which is not consistent with the [official tutorial](69b563dc3b0ea293d5c3a0a89343c75b285983fb)

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/hooks/profiler_hook.py
+++ b/mmengine/hooks/profiler_hook.py
@@ -136,7 +136,6 @@ class ProfilerHook(Hook):
 
         self.json_trace_path = json_trace_path
 
-    @master_only
     def before_run(self, runner):
         """Initialize the profiler.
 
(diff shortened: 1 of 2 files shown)


BayMaxBHL (approved):

mmengine/hooks/profiler_hook.py
```diff
@@ -48,12 +48,16 @@ class ProfilerHook(Hook):
...
+            Two officially recommended ways are provided:
+
+            - ``schedule=dict(type='log_trace')``: Print the profiling result
+              in the terminal See more details in the `official tutorial`_.
+              The configurable arguments is the same as
+              ``prof.key_averages().table``
+            - ``scheduler=dict(type='tb_trace')``: Profile the performance
+               with tensorboard. See more details in the tutorial
+               `profile with tensorboard`_.
+
```
zhouzaida, line 60:
```suggestion
            Two officially recommended ways are provided:

            - ``schedule=dict(type='log_trace')``: Print the profiling result
              in the terminal. See more details in the `PyTorch official tutorial`_.
              The configurable arguments are the same as
              ``prof.key_averages().table``
            - ``scheduler=dict(type='tb_trace')``: Profile the performance
              with tensorboard. See more details in the tutorial
              `profile with tensorboard`_.

```

zhouzaida (dismissed):

PR #1152, opened 2023-05-17
[Fix] Correctly record random seed in log

Shiyang980713:
## Motivation

When no random seed is specified, the random seed value is recorded as None in the log

## Modification

Assign self._seed to self._randomness_cfg['seed']

### DIFF
--- a/mmengine/runner/runner.py
+++ b/mmengine/runner/runner.py
@@ -355,6 +355,7 @@ class Runner:
         # `set_randomness`` method
         self._randomness_cfg = randomness
         self.set_randomness(**randomness)
+        self._randomness_cfg['seed'] = self._seed
 
         if experiment_name is not None:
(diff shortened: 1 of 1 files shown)


mmengine/runner/runner.py
```diff
@@ -355,6 +355,7 @@ def __init__(
         # `set_randomness`` method
         self._randomness_cfg = randomness
         self.set_randomness(**randomness)
+        self._randomness_cfg['seed'] = self._seed
```
zhouzaida, line 358: It's not recommended to modify `self._randomness_cfg`, just update it when logging the environment. https://github.com/open-mmlab/mmengine/blob/ec2e00ae9060572de96759ef141254d352134d39/mmengine/runner/runner.py#L2308

```python
runtime_env.update(self._randomness_cfg)
runtime_env['seed'] = self._seed
```

zhouzaida (approved):

PR #1163, opened 2023-05-25
[Fix] Fix unit test failed for limited CI resources

HAOCHENYE:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/testing/_internal/distributed.py
+++ b/mmengine/testing/_internal/distributed.py
@@ -1,10 +1,10 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 # Copyright (c) https://github.com/pytorch/pytorch
 # Modified from https://github.com/pytorch/pytorch/blob/master/torch/testing/_internal/common_distributed.py  # noqa: E501
-
 import faulthandler
 import logging
 import multiprocessing
+import signal
 import sys
 import tempfile
 import threading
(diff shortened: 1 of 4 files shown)


mmengine/testing/_internal/distributed.py
```diff
@@ -172,7 +171,10 @@ def _event_listener(parent_pipe, signal_pipe, rank: int):
     def _run(cls, rank: int, test_name: str, file_name: str,
              parent_pipe) -> None:
         self = cls(test_name)
-        self.prepare_subprocess()
+        try:
+            self.prepare_subprocess()
```
zhouzaida, line 175: Why might this method throw an exception?

HAOCHENYE: To throw an exit code to help `unittest` or `pytest` catch the exception and raise an error. 

Actually, previous `MultiProcessTest` will fail in many cases:

1. Fail to initialize the subprocess for the limited resources, raise an unexpected error will be rasied.  In this case, the error will be raised be for executing the `tested method`.
2. The system terminates the subprocess for some "unknown" reason, maybe system overload.
3. The tested function failed
4. The `prepare_subprocess` failed.

However, CI fails for reasons 1 and 2 mostly. This PR will skip the unit test for reason 1 and reason 2, and keep others unchanged

PR #1168, opened 2023-05-26
[Fix] Update the best_ckpt of checkpoint if the current epoch has the best performance

zhouzaida:

## Motivation

save checkpoint again to update the best_score and best_ckpt stored in message_hub because the checkpoint saved in `after_train_epoch` or `after_train_iter` stage only keeps the previous best checkpoint not the current best checkpoint which causes the current best checkpoint can not to be removed when resuming training.

## Modification

- save checkpoint again to update best_ckpt of ckpt

- If the current checkpoint is the best checkpoint, it will be saved without the effect of interval.

  For example, if interval=2 and the current epoch is 5, only one model weight without optimizer state will be saved, but now the checkpoint with optimizer state will also be saved.

- update the logic to remove outdated checkpoints

  For the same reason as above, these checkpoints that will be deleted are somewhat different from the previous behaviour (before this PR). For example, if interval=2 and the 5th epcoh is the best checkpoint, epoch_2.pth, epoch_4.pth will be kept, and now there will be one more epoch_5.pth. Supposing max_keep_ckpts=1, only epoch_4.pth will be kept before, and now only epoch_4.pth will be kept. I think this is an acceptable change.

- save keep_ckpt_ids to ckpt for resuming training

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/tests/test_hooks/test_checkpoint_hook.py
+++ b/tests/test_hooks/test_checkpoint_hook.py
@@ -499,11 +499,13 @@ class TestCheckpointHook(RunnerTestCase):
         runner = self.build_runner(cfg)
         runner.train()
         self.assertTrue(
-            osp.isfile(osp.join(cfg.work_dir, f'{training_type}_10.pth')))
+            osp.isfile(osp.join(cfg.work_dir, f'{training_type}_11.pth')))
 
-        for i in range(10):
+        for i in range(11):
             self.assertFalse(
                 osp.isfile(osp.join(cfg.work_dir, f'{training_type}_{i}.pth')))
+
+        cfg.default_hooks.checkpoint.max_keep_ckpts = -1
 
         # Test filename_tmpl
         cfg.default_hooks.checkpoint.filename_tmpl = 'test_{}.pth'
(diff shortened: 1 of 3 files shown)


tests/test_hooks/test_checkpoint_hook.py
```diff
@@ -539,8 +576,46 @@ def test_with_runner(self, training_type):
...
+        ckpt = torch.load(osp.join(cfg.work_dir, f'{training_type}_5.pth'))
+        self.assertEqual(best_ckpt,
+                         ckpt['message_hub']['runtime_info']['best_ckpt'])
+
+        if training_type == 'epoch':
+            self.assertEqual(ckpt['meta']['epoch'], 5)
+            self.assertEqual(ckpt['meta']['iter'], 20)
+        else:
+            self.assertEqual(ckpt['meta']['epoch'], 0)
+            self.assertEqual(ckpt['meta']['iter'], 5)
```
HAOCHENYE, line 588: Is it necessary to check these intermediate variable here?

zhouzaida: Those variables may be incorrect if they are not handled carefully.

HAOCHENYE (approved):

PR #1181, opened 2023-06-01
Refine the error message when auto_scale_lr is not set correctly

alexander-soare:
## Motivation

RuntimeError when attempting to resume a run. It seems the logic for triggering the runtime error is incorrect (or the error message is incorrect, but this PR is based on the assumption that the error message is correct).

## Modification

Remove a `not` to fix the logic. This makes the RuntimeError trigger if `self.auto_scale_lr` has a key called "enable" and that key is `True`. This is the desired logic if we consider the error message which asks the user to set "enable" to `False` to avoid the error.

## Checklist

I did this in 30 seconds so admittedly I haven't followed the checklist. I'd be happy to file an issue and allow a core maintainer to carry on from here. Please let me know.

### DIFF
--- a/mmengine/runner/runner.py
+++ b/mmengine/runner/runner.py
@@ -1978,11 +1978,10 @@ class Runner:
                 if (self.auto_scale_lr is None
                         or self.auto_scale_lr.get('enable', False)):
                     raise RuntimeError(
-                        'Number of GPU used for current experiment is not '
-                        'consistent with resuming from checkpoint, '
-                        'which may result in poor performance due to the '
-                        'learning rate. If you want the learning rate to '
-                        'be automatically adjusted, set the '
+                        'Number of GPUs used for current experiment is not '
+                        'consistent with the checkpoint being resumed from. '
+                        'This will result in poor performance due to the '
+                        'learning rate. You must set the '
                         '`auto_scale_lr` parameter for Runner and make '
                         '`auto_scale_lr["enable"]=True`.')
                 else:
--- a/mmengine/runner/runner.py
+++ b/mmengine/runner/runner.py
@@ -1975,16 +1975,22 @@ class Runner:
             if (previous_gpu_ids is not None and len(previous_gpu_ids) > 0
                     and len(previous_gpu_ids) != self._world_size):
                 # TODO, should we modify the iteration?
-                self.logger.info(
-                    'Number of GPU used for current experiment is not '
-                    'consistent with resuming from checkpoint')
                 if (self.auto_scale_lr is None
                         or self.auto_scale_lr.get('enable', False)):
                     raise RuntimeError(
-                        'Cannot automatically rescale lr in resuming. Please '
-                        'make sure the number of GPU is consistent with the '
-                        'previous training state resuming from the checkpoint '
-                        'or set `enable` in `auto_scale_lr` to False.')
+                         'Number of GPU used for current experiment is not '
+                         'consistent with resuming from checkpoint, '
+                         'which may result in poor performance due to the '
+                         'learning rate. If you want the learning rate to '
+                         'be automatically adjusted, set the '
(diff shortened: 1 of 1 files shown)


zhouzaida (dismissed):

mmengine/runner/runner.py
```diff
@@ -1979,12 +1979,12 @@ def resume(self,
                     'Number of GPU used for current experiment is not '
                     'consistent with resuming from checkpoint')
                 if (self.auto_scale_lr is None
-                        or not self.auto_scale_lr.get('enable', False)):
+                        or self.auto_scale_lr.get('enable', False)):
```
zhouzaida, line 1982: Hi @alexander-soare , I think we should keep the `not` and modify the error message from "set `enable` in `auto_scale_lr` to False" to "set `enable` in `auto_scale_lr` to True".

alexander-soare: Sure, but then I think the rest of the error message needs more clarity. It begins with "Cannot automatically rescale lr in resuming." That sentence alone implies that `auto_scale_lr` should be `False`. And apologies: I'm not aware of the underlying logic as I'm new to MMDet, so I can't suggest a good error message myself.

zhouzaida: I will try to propose an error message.

mmengine/runner/runner.py
```diff
@@ -1975,16 +1975,22 @@ def resume(self,
...
-                        'make sure the number of GPU is consistent with the '
-                        'previous training state resuming from the checkpoint '
-                        'or set `enable` in `auto_scale_lr to False.')
+                        'Number of GPU used for current experiment is not '
+                        'consistent with resuming from checkpoint, '
+                        'which may result in poor performance due to the '
+                        'learning rate. If you want the learning rate to '
+                        'be automatically adjusted, set the '
+                        '`auto_scale_lr` parameter for Runner and make '
+                        '`auto_scale_lr["enable"]=True`.')
```
alexander-soare, line 1987: This suggestion incorporates some small aesthetic tweaks but importantly it changes the language to be more imperative (please consider my comment above).

```suggestion
                        'Number of GPUs used for current experiment is not '
                        'consistent with the checkpoint being resumed from. '
                        'This will result in poor performance due to the '
                        'learning rate. You must set the '
                        '`auto_scale_lr` parameter for Runner and make '
                        '`auto_scale_lr["enable"]=True`.')
```

mmengine/runner/runner.py
alexander-soare: Nice. I think that's better. One nit though: the use of the language "If you want" suggests this is optional. But the RuntimeError leaves the user with no choice. I would suggest you either change the word choice (see my suggestion below) or you make this a warning rather than an exception.

zhouzaida: 👍I think your refinement makes it better and I will merge this PR.

mmengine/runner/runner.py
```diff
@@ -1975,16 +1975,21 @@ def resume(self,
             if (previous_gpu_ids is not None and len(previous_gpu_ids) > 0
                     and len(previous_gpu_ids) != self._world_size):
                 # TODO, should we modify the iteration?
-                self.logger.info(
-                    'Number of GPU used for current experiment is not '
-                    'consistent with resuming from checkpoint')
                 if (self.auto_scale_lr is None
-                        or not self.auto_scale_lr.get('enable', False)):
+                        or self.auto_scale_lr.get('enable', False)):
```
zhouzaida, line 1979:
```suggestion
                        or not self.auto_scale_lr.get('enable', False)):
```

zhouzaida (approved):

mmengine/runner/runner.py
alexander-soare: @zhouzaida thanks for the clarification!

PR #977, opened 2023-03-04
[Enhance] WandbVisBackend supports updating config

zgzhengSEU:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

Please describe the motivation of this PR and the goal you want to achieve through this PR.

support wandb resume

## Modification

Please briefly describe what modification is made in this PR.

support wandb resume

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repos?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #1185, opened 2023-06-02
Add allow_config_update option

alexander-soare:
## Motivation

Sometimes one wishes to resume runs with WandB enabled. But for resuming runs sometimes it is necessary to change some parameters in the MMDetection config. For example, on first attempting a run, one might set `load_from`. But then when resuming the run, `load_from` should be set to `None` to let the resumer pick the latest checkpoint. `wandb.config.update` then raises a `ConfigError` because we are attempting to pass in a different config key. WandB offers a `allow_val_change` option but this is not exposed by the MMDetection API. (note: it's possible to pass `allow_val_change` as `init_kwargs` for WandB but this is a **different** parameter to the one used in `wandb.config.update`)

## Modification

Adds and option to `WandbVisBackend` called `allow_config_update`. Adds a private attribute `_allow_config_update`. Passes `_allow_config_update` to `wandb.config.update`.

## BC-breaking (Optional)

Not BC-breaking.

### DIFF
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -347,6 +347,10 @@ class WandbVisBackend(BaseVisBackend):
             input parameters.
             See `wandb.init <https://docs.wandb.ai/ref/python/init>`_ for
             details. Defaults to None.
+        allow_config_update (bool, optional): Uses this value for
+            `allow_val_change` in the call to `wandb.config.update`. One may
+            wish to set this to `True` if the are resuming runs and have any
+            changes to the MMDetection config.
         define_metric_cfg (dict or list[dict], optional):
             When a dict is set, it is a dict of metrics and summary for
             ``wandb.define_metric``.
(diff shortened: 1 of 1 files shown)


zhouzaida:
Hi @alexander-soare , thanks for your contribution. There is a same PR in #977.

alexander-soare:
@zhouzaida thanks for letting me know. I suppose the difference between this PR and that one is that this one allows that keyword to remain independent for both the init and the update. That one couples the two. I can't personally think of any advantages of decoupling them as I have done, but I don't use WandB often enough to know for sure if it's the right choice. Happy to leave it up to the core maintainers.

zhouzaida:
> @zhouzaida thanks for letting me know. I suppose the difference between this PR and that one is that this one allows that keyword to remain independent for both the init and the update. That one couples the two. I can't personally think of any advantages of decoupling them as I have done, but I don't use WandB often enough to know for sure if it's the right choice. Happy to leave it up to the core maintainers.

Thank you for your contribution. I lean towards the solution proposed in #977 as well because it avoids adding additional parameters to WandbVisBackend.

PR #1188, opened 2023-06-03
[Feature] Support using other file handlers

KevinNuNu:

## Motivation

在应用上线后不能让日志文件大小无限增大，需要对日志按要求进行分割。

## Modification

为MMLogger增加一个接口使得支持日志分割。

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


mmengine/logging/logger.py
```diff
@@ -170,6 +171,19 @@ class MMLogger(Logger, ManagerMixin):
         log_level (str): The log level of the handler. Defaults to
             'INFO'. If log level is 'DEBUG', distributed logs will be saved
             during distributed training.
+        log_rotating (dict, optional): The cfg dict of file handler that
```
HAOCHENYE, line 174: Suggest renaming `log_rotating` to `file_handler`. If it is None, `__init__` will initialize it as `dict(type='FileHandler')`. We can also maintain a private registry here, for examples:

```python
from mmengine.registry import Registry

LOGGER_HANDLERS = Registry('logger handler')
# register handlers in logging to `LOGGER_HANDLERS`
```

And then build the file handler from `LOGGER_HANDLERS.build(file_handler)`

KevinNuNu: I have renamed log_rotating to file_handler_cfg.

HAOCHENYE: Hi, you can refer to the implementation here:

https://github.com/open-mmlab/mmengine/blob/1c3332c96e9ef0a591850b2c536d9a5fc28f11b7/mmengine/dataset/utils.py#L16

Add the following code

```python
FILE_HANDLERS = Registry('file handler')
```

at the top of the `logger.py`

zhouzaida (approved):

PR #1202, opened 2023-06-15
[Feature] Turning On fast_conv_bn_eval

youkaichao:
## Motivation

The `mmcv` package will introduce a `fast_conv_bn_eval` feature in `2.0.1`, as discussed in https://github.com/open-mmlab/mmcv/pull/2807 .

This PR enables the `mmengine` to use that feature.

## Modification

`Runner` tries to read from `cfg` to know if the `fast_conv_bn_eval` option is turned on. If it is, and the `mmcv` supports `fast_conv_bn_eval`, then:

- `fast_conv_bn_eval` feature will be turned on for existing `ConvModule`.
- Consecutive conv + bn blocks will be merged into a new `ConvModule` (with existing parameters).

## BC-breaking (Optional)

This feature does not break the backward-compatibility. It relies on a new cfg option `fast_conv_bn_eval`.

## Use cases (Optional)

Downstream projects can leverage this feature by adding command-line options `fast_conv_bn_eval=True`. They also need to specify the parts of module to use with this features. For example, `fast_conv_bn_eval_modules=['backbone']`. This limitation comes from the fact that whole models might be too complex to trace (e.g. detection models) to find consecutive conv-bn blocks, while parts of the model are simple enough to trace. The main usecases are to replace consecutive conv-bn blocks in the pre-trained backbone.

## Checklist

- [x] Pre-commit or other linting tools are used to fix the potential lint issues.
- [ ]  The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [x]  If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
- [x]  The documentation has been modified accordingly, like docstring or example tutorials.


youkaichao:
Meeting Notes with @zhouzaida  and @HAOCHENYE :

The logic happens in `Runner.train`, and the order of `build_optim_wrapper` / `_init_model_weights` / `fast_conv_bn_eval` matters.

Ideally, we would like to have `_init_model_weights` --> `fast_conv_bn_eval`  --> `build_optim_wrapper`. However, this is not available since we have `build_optim_wrapper` --> `_init_model_weights` now.

Some users would set param groups when building optimizers. The param groups might depend on the key values of parameters. To keep these users unaffected, we should place `fast_conv_bn_eval` after `build_optim_wrapper`.

Therfore, we finally agreed on the order of `build_optim_wrapper` --> `_init_model_weights` --> `fast_conv_bn_eval`.

youkaichao:
Tested on MMDetection, with a small fraction of coco data, in my laptop. The optimizer works fine.

docs/zh_cn/common_usage/save_gpu_memory.md
```diff
@@ -2,6 +2,10 @@
 
 在深度学习训练推理过程中显存容量至关重要，其决定了模型是否能成功运行。常见的节省显存办法包括：
 
+- 启用快速卷积BN评估功能（实验性）
+
+  基于在[这篇论文](https://arxiv.org/abs/2305.11624)中讨论的概念，我们最近在MMCV中[引入](https://github.com/open-mmlab/mmcv/pull/2807)了一个实验性功能：快速卷积BN评估。这个功能的设计目标是在不损害性能的情况下减少网络训练过程中的显存占用。如果你的网络架构包含了一系列连续的Conv+BN模块，而且这些BN层在训练过程中保持在 `eval` 模式（在使用 [MMDetection](https://github.com/open-mmlab/mmdetection)训练对象检测器时很常见），这个功能可以将显存消耗减少超过 $20%$。要启用快速卷积BN评估功能，只需添加以下命令行参数：`--cfg-options fast_conv_bn_eval=True fast_conv_bn_eval_modules="[backbone]"`。当你在输出日志中看到 `Enabling the "fast_conv_bn_eval" feature for these modules ...`时，意味着功能已成功启用。由于这仍处于实验阶段，我们非常期待听到你对它的使用体验。请在[这个讨论线程](https://github.com/open-mmlab/mmcv/discussions/2841)分享你的使用报告、观察和建议。你的反馈对于进一步的开发和确定是否应将此功能集成到稳定版中至关重要。
```
HAOCHENYE, line 7: It could be better to merge `fast_conv_bn_eval_modules` and `fast_conv_bn_eval` into one argument.

HAOCHENYE (commented):
Hi~ We should provide corresponding unittest for `turn_on_fast_conv_bn_eval` to check output will not be influenced when model is modified by it. It could be more reliable to validate that `turn_on_fast_conv_bn_eval` works fine with different network architecture.

HAOCHENYE:
BTW, we should also update this function in [FlexibleRunner](https://github.com/open-mmlab/mmengine/blob/main/mmengine/runner/_flexible_runner.py)

mmengine/runner/runner.py
```diff
@@ -1705,6 +1707,34 @@ def train(self) -> nn.Module:
...
         self._init_model_weights()
+
+        # try to enable fast_conv_bn_eval feature
+        fast_conv_bn_eval = self.cfg.get('fast_conv_bn_eval', False)
+        if fast_conv_bn_eval:
+            can_use_fast_conv_bn_eval = False
+            if is_installed('mmcv'):
+                from mmcv.cnn import ConvModule
+                if hasattr(ConvModule, 'turn_on_fast_conv_bn_eval'):
+                    can_use_fast_conv_bn_eval = True
```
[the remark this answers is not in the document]
youkaichao: An error is raised now, but in the code below. Resolved by  d9fe010 .

docs/zh_cn/common_usage/save_gpu_memory.md
youkaichao: Resolved in [e3ddabc](https://github.com/open-mmlab/mmengine/pull/1202/commits/e3ddabc8c1ec1e125dc23d874691eaa3622c5dde).

mmengine/model/turn_on_fast_conv_bn_eval.py
```diff
@@ -0,0 +1,73 @@
...
+    for node in fx_model.graph.nodes:
+        # If our current node isn't calling a Module then we can ignore it.
+        if node.op != 'call_module':
+            continue
+        found_pair = [
+            node for conv_class, bn_class in patterns
+            if isinstance(modules[node.target], bn_class)
+            and isinstance(modules[node.args[0].target], conv_class)
+        ]
+        # Not a conv-BN pattern or output of conv is used by other nodes
```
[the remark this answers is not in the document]
youkaichao: Fixed in [804c7fd](https://github.com/open-mmlab/mmengine/pull/1202/commits/804c7fd212f77b85f4dba5c0171325ee43fb84b7).

mmengine/model/turn_on_fast_conv_bn_eval.py
```diff
@@ -0,0 +1,73 @@
...
+        if not found_pair or len(node.args[0].users) > 1:
```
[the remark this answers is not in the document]
youkaichao: In your case, I would suggest to use the forward function as:

```python
class ToyModel(nn.Module):
    def __init__(self, *args, **kwargs) -> None:
        super().__init__(*args, **kwargs)
        self.conv1 = nn.Conv2d(1, 1, 1)
        self.bn1 = nn.BatchNorm2d(1)
    
    def forward(self, x):
        x = self.conv1(x)
        return x + self.bn1(x)
```

Then the previous implementation works, too.

Nevertheless, you raised a valid concern. This is fixed in [af318cc](https://github.com/open-mmlab/mmengine/pull/1202/commits/af318ccd460f7724abdbe2eeb3dd4cbb3e5cbeb1).

youkaichao:
> Hi~ We should provide corresponding unittest for `turn_on_fast_conv_bn_eval` to check output will not be influenced when model is modified by it. It could be more reliable to validate that `turn_on_fast_conv_bn_eval` works fine with different network architecture.

Unit test is intended to be used for simple functions. However, the validity of this feature usually requires days of GPUs to run detection experiments on the whole MSCOCO dataset. Do we have this much resource for unit testing?

HAOCHENYE:
> > Hi~ We should provide corresponding unittest for `turn_on_fast_conv_bn_eval` to check output will not be influenced when model is modified by it. It could be more reliable to validate that `turn_on_fast_conv_bn_eval` works fine with different network architecture.
> 
> Unit test is intended to be used for simple functions. However, the validity of this feature usually requires days of GPUs to run detection experiments on the whole MSCOCO dataset. Do we have this much resource for unit testing?

We don't need to validate the training accuracy for the specific model. We can construct some `ToyModel` with different `Conv` + `BN` structures, and then validate whether the output of the modified `ToyModel` instance matches with the prior one.

youkaichao:
> > > Hi~ We should provide corresponding unittest for `turn_on_fast_conv_bn_eval` to check output will not be influenced when model is modified by it. It could be more reliable to validate that `turn_on_fast_conv_bn_eval` works fine with different network architecture.
> > 
> > 
> > Unit test is intended to be used for simple functions. However, the validity of this feature usually requires days of GPUs to run detection experiments on the whole MSCOCO dataset. Do we have this much resource for unit testing?
> 
> We don't need to validate the training accuracy for the specific model. We can construct some `ToyModel` with different `Conv` + `BN` structures, and then validate whether the output of the modified `ToyModel` instance matches with the prior one. The structure should include the cases that all `BN` and `Conv` can be merged, only part of pairs can be merged, and none of the submodules can be merged.

Fixed in [23bc688](https://github.com/open-mmlab/mmengine/pull/1202/commits/23bc688da399531e53a1f77ed3560a8dd9f87879).

mmengine/model/turn_on_fast_conv_bn_eval.py
zhouzaida, line 1: simplify the filename `turn_on_fast_conv_bn_eval.py`. suggest renaming it to `fast_conv_bn_eval.py`.

youkaichao: Fixed in [30622cd](https://github.com/open-mmlab/mmengine/pull/1202/commits/30622cdd0c13c96f4f18c4082d89feb25cd73f26).

mmengine/runner/runner.py
```diff
@@ -1705,6 +1705,19 @@ def train(self) -> nn.Module:
...
         self._init_model_weights()
+
+        # try to enable fast_conv_bn_eval feature
+        fast_conv_bn_eval = self.cfg.get('fast_conv_bn_eval', None)
+        if fast_conv_bn_eval is not None:
+            from mmengine.model.fast_conv_bn_eval import \
+                turn_on_fast_conv_bn_eval
+            modules = fast_conv_bn_eval
+            if isinstance(modules, str):
+                modules = [modules]
```
zhouzaida, line 1716: These lines can be also moved to `turn_on_fast_conv_bn_eval`?

mmengine/runner/runner.py
```diff
@@ -1705,6 +1705,19 @@ def train(self) -> nn.Module:
 
         # initialize the model weights
         self._init_model_weights()
+
+        # try to enable fast_conv_bn_eval feature
+        fast_conv_bn_eval = self.cfg.get('fast_conv_bn_eval', None)
+        if fast_conv_bn_eval is not None:
+            from mmengine.model.fast_conv_bn_eval import \
+                turn_on_fast_conv_bn_eval
```
zhouzaida, line 1713: This import statement can be moved to the head of this file.

tests/test_model/test_fast_conv_bn_eval.py
```diff
@@ -0,0 +1,52 @@
...
+        # this conv-bn pair cannot use fast_conv_bn_eval feature
+        # because `self.conv2` is used twice
+        x = self.bn2(self.conv2(self.conv2(x)))
+        # this conv-bn pair cannot use fast_conv_bn_eval feature
+        # because `self.bn3` is used twice
+        x = self.bn3(self.bn3(self.conv3(x)))
+        return x
+
+
+class TestFastConvBNEval(TestCase):
```
zhouzaida, line 41: torch.fx is a new feature of torch1.8+ so it is necessary to skip unit tests when torch<1.8.0. Refer to https://github.com/open-mmlab/mmengine/blob/b2295a258c4af3cdeb1a55fa5e49cddbd1c6c0bf/tests/test_model/test_wrappers/test_model_wrapper.py#L224

mmengine/runner/runner.py
youkaichao: This import statement might raise RuntimeError. It is safer to keep it here to minimize the influence of code not turning on `fast_conv_bn_eval` feature.

mmengine/runner/runner.py
youkaichao: Fixed in [4e434fb](https://github.com/open-mmlab/mmengine/pull/1202/commits/4e434fb765099c398eb8cc5991d444d90b39753f).

mmengine/model/fast_conv_bn_eval.py
```diff
@@ -0,0 +1,103 @@
...
+    return parent[0] if parent else '', name
+
+
+def replace_sub_module(model, name, new_module):
+    # Remove the original module from the model
+    # usage: replace_sub_module(model, 'layer1.block2.conv2', conv)
+    parent_name, name = _parent_name(name)
+    if parent_name != '':
+        getter = attrgetter(parent_name)
+        parent = getter(model)
```
zhouzaida, line 32: Can we directly use `getattr` rather than the two lines?

youkaichao: Fixed in [0de92fa](https://github.com/open-mmlab/mmengine/pull/1202/commits/0de92fa65e066592ace48491a31a94d4936d783d). Note that `getattr` does not work here. The `parent_name` might have multiple levels like `layer1.bottleneck`. Therefore, we need  `attrgetter`.

mmengine/model/fast_conv_bn_eval.py
```diff
@@ -0,0 +1,176 @@
...
+
+        conv_module.forward = partial(fast_conv_bn_eval_control, bn_module,
+                                      conv_module)
+
+
+def turn_on_fast_conv_bn_eval(model: torch.nn.Module, modules: Union[List[str],
+                                                                     str]):
+    if isinstance(modules, str):
+        modules = [modules]
+    global optimize_conv_module
```
zhouzaida, line 172:
```suggestion
```

zhouzaida: This line can be removed.

tests/test_model/test_fast_conv_bn_eval.py
```diff
@@ -0,0 +1,59 @@
...
+from mmengine.testing import assert_allclose
+
+optimize_conv_module = mod.optimize_conv_module
+
+
+class BackboneModel(nn.Module):
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        global optimize_conv_module
```
zhouzaida, line 19:
```suggestion
```

zhouzaida: This line can be removed.

tests/test_model/test_fast_conv_bn_eval.py
```diff
@@ -0,0 +1,59 @@
...
+            self.mod1 = ConvModule.create_from_conv_bn(conv0, bn0)
+        self.conv1 = nn.Conv2d(6, 6, 6)
+        self.bn1 = nn.BatchNorm2d(6)
+        self.conv2 = nn.Conv2d(6, 6, 6)
+        self.bn2 = nn.BatchNorm2d(6)
+        self.conv3 = nn.Conv2d(6, 6, 6)
+        self.bn3 = nn.BatchNorm2d(6)
+
+    def forward(self, x):
+        global optimize_conv_module
```
zhouzaida, line 33: same as above

youkaichao: My bad. Now I remember that `global` is used to modify global variables. These lines can be removed. Fixed in [bee4711](https://github.com/open-mmlab/mmengine/pull/1202/commits/bee47111a16e086b2e666c3868c25a6262cf1d79).

youkaichao:
This PR enables `fast_conv_bn_eval` feature for pytorch native `conv` and `bn`. It works for all the modules whose computation graphs does not change during training.

This PR does not support dynamic computation graph (which changes across different `forward` calls). A possible example of dynamic computation graph is [`ConvModule`](https://mmcv.readthedocs.io/en/latest/_modules/mmcv/cnn/bricks/conv_module.html), whose `forward` function can take a `norm` argument to control whether to use normalization layers for each `forward` pass. However, the `ConvModule` itself in mmcv does support `fast_conv_bn_eval` feature even in the case of dynamic computation graph. If users want to use `fast_conv_bn_eval` with `ConvModule` and a dynamic `norm` argument (a rather rare usecase), then they can depend on `ConvModule. turn_on_fast_conv_bn_eval` rather than `from mmengine.model.fast_conv_bn_eval import turn_on_fast_conv_bn_eval.`

To sum up:
- If you don't use `ConvModule`, this PR is enough to use `fast_conv_bn_eval` feature.
- If you use `ConvModule` and the `norm` argument keeps the same across `forward` calls, this PR is enough to use `fast_conv_bn_eval` feature.
- If you use `ConvModule` and **the `norm` argument changes across `forward` calls**, this might fail `torch.fx.symbolic_trace`. Please do not use this PR. Instead, use `ConvModule. turn_on_fast_conv_bn_eval` function manually for each `ConvModule`.

zhouzaida (approved):

HAOCHENYE (approved):

PR #1239, opened 2023-07-06
[Fix] Fix the logic of setting lazy_import

Li-Qingyun:
## Failure case:

a config which is allowed by previous mmengine:
```
from mmdet.datasets import CocoDataset
... = dict(
    ...=dict(
        class_names_txt=CocoDataset.METAINFO['classes']
    )
)

# delete python style config to allow printing and dumping config
del CocoDataset
```
This is non-lazy_import, hence i provide a hack arg in train.py:
```
cfg = Config.from_file(arg.cfg, lazy_import=False)
```
But there is still auto lazy_import detected:
ad3c410a64868dec60df6cc26d9cfd3

I find auto detecting lazy_import maintains in `_file2dict()`.
But **If a user provide `lazy_import=False`, why auto detecting lazy_import again? If whether lazy_import is True or False, it must be equal to `Config._is_lazy_import(filename)`, what is the meaning of `lazy_import=False` arg.**

## Solutions:
1. also add `lazy_import` arg to `_file2dict()`    (but `_file2dict()` seems customized for non-lazy_import, seems not pretty)
2. delete the auto detecting       (but may raise anothor error)
3. add an `self.lazy_import` attribute.


tests/data/config/lazy_module_config/error_mix_using3.py
```diff
@@ -0,0 +1,2 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import numpy as np
```
zhouzaida, line 2:
```suggestion
import numpy as np

```

mmengine/config/config.py
```diff
@@ -421,11 +421,11 @@ def fromfile(filename: Union[str, Path],
             Config: Config instance built from config file.
         """
         filename = str(filename) if isinstance(filename, Path) else filename
-        if lazy_import is None:
-            lazy_import = Config._is_lazy_import(filename)
-        if not lazy_import:
+        inferred_lazy_import = Config._is_lazy_import(filename)
+        if not inferred_lazy_import or lazy_import is False:
```
zhouzaida, line 425:
```suggestion
        if (lazy_import is False or
            (lazy_import is None and not Config._is_lazy_import(filename))):
```

zhouzaida (dismissed):

PR #1250, opened 2023-07-14
Fix scalar check in RuntimeInfoHook

i-aki-y:
## Motivation

Some numpy scalars (float32, float16, ...) are misclassified as non-scalar values inRuntimeInfoHook.

See:

```
from mmengine.hooks.runtime_info_hook import _is_scalar
v = np.mean([0.5])
print(_is_scalar(v.astype(np.float16)))  # => False
print(_is_scalar(v.astype(np.float32)))  # => False
print(_is_scalar(v.astype(np.float64)))  # => True
```

This causes problems when the custom metrics are calculated in numpy because the log_processor looks to collect only scalar values as metrics to give the visualizer them.

## Modification

Added np.floating and np.integer as valid scalar types.

## BC-breaking (Optional)

## Use cases (Optional)

I encountered this problem when I defined custom metrics where I calculated metrics in tensor and converted the result to numpy. (ex. metrics.detach().cpu().numpy() --> np.float32)

## Checklist

### DIFF
--- a/mmengine/hooks/runtime_info_hook.py
+++ b/mmengine/hooks/runtime_info_hook.py
@@ -23,7 +23,7 @@ def _is_scalar(value: Any) -> bool:
     """
     if isinstance(value, np.ndarray):
         return value.size == 1
-    elif isinstance(value, (int, float, np.floating, np.integer)):
+    elif isinstance(value, (int, float, np.number)):
         return True
     elif isinstance(value, torch.Tensor):
         return value.numel() == 1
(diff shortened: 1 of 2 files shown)


mmengine/hooks/runtime_info_hook.py
```diff
@@ -23,7 +23,7 @@ def _is_scalar(value: Any) -> bool:
     """
     if isinstance(value, np.ndarray):
         return value.size == 1
-    elif isinstance(value, (int, float)):
+    elif isinstance(value, (int, float, np.floating, np.integer)):
```
zhouzaida, line 26: Can we replace `np.floating, np.integer` with `np.number`?

i-aki-y: @zhouzaida  Thank you for your comment. Yes, I missed it. 
The np.number would be better.

For clarification, a difference is that np.number includes complex numbers while the `np.floating, np.integer` does not.
https://numpy.org/doc/stable/reference/arrays.scalars.html

zhouzaida (approved):

Issue #1076, opened 2023-04-14
[Feature] Support comparing two `Config`.


HAOCHENYE:
### What is the feature?

MMEngine should add a function to compare

```python
cfg1 = Config.fromfile('cfg1')     #  cfg1 = dict(a=1, b=2)
cfg2 = Config.fromfile('cfg2')     #  #  cfg1 = dict(a=1, b=3)

Config.diff(cfg1, cfg2)
```

The result should be:
```bash
dict(a=1, b=2)  # the b=2 should be red
```

IRONICBo:
Is the result returned by diff based on cfg1 or cfg2? For example, if cfg1 is a subset of cfg2, if cfg1 is returned, cfg1 may not show the difference in red above.

For example cfg1=dict(a=1), cfg2=dict(a=1,b=2)

HAOCHENYE:
The formatting of the diff results requires further refinement and design. The example provided in the issue should not be taken as the final design.

In my opinion, the format of the `git diff` results is understandable and useful :rofl: .

gachiemchiep:
@HAOCHENYE 

Using python's difflib library, I can generate a diff as git style.  

* command

```python
cfg1 = Config.fromfile('config1.py')     #  cfg1 = dict(a=1, b=2)
cfg2 = Config.fromfile('config2.py')     #  #  cfg1 = dict(a=1, b=3)
print(Config.diff(cfg1, cfg2))
```

* output

```bash
--- 

+++ 

@@ -12,7 +12,7 @@

         num_classes=10,
         in_channels=512,
         loss=dict(type='CrossEntropyLoss', loss_weight=1.0)))
-dataset_type = 'CIFAR10'
+dataset_type = 'CIFAR100'
 data_preprocessor = dict(
     num_classes=10,
     mean=[
@@ -38,7 +38,7 @@

     batch_size=16,
     num_workers=2,
     dataset=dict(
-        type='CIFAR10',
+        type='CIFAR100',
         data_prefix='data/cifar10',
         test_mode=False,
         pipeline=[
@@ -51,7 +51,7 @@

     batch_size=16,
     num_workers=2,
     dataset=dict(
-        type='CIFAR10',
+        type='CIFAR100',
         data_prefix='data/cifar10/',
         test_mode=True,
         pipeline=[
@@ -63,7 +63,7 @@

     batch_size=16,
     num_workers=2,
     dataset=dict(
-        type='CIFAR10',
+        type='CIFAR100',
[...]

PR #1260, opened 2023-07-19
[Feature] Compare the difference of two configs

gachiemchiep:

## Motivation

As discussion in : https://github.com/open-mmlab/mmengine/issues/1076

## Modification

Add diff method to compare 2 configuration 

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


HAOCHENYE:
Hi, it could be helpful to include a unit test [here](https://github.com/open-mmlab/mmengine/blob/main/tests/test_config/test_config.py) :smile:

gachiemchiep:
@HAOCHENYE unit test added

mmengine/config/config.py
```diff
@@ -1605,6 +1606,12 @@ def merge_from_dict(self,
             Config._merge_a_into_b(
                 option_cfg_dict, cfg_dict, allow_list_keys=allow_list_keys))
 
+    @staticmethod
+    def diff(cfg1, cfg2) -> str:
```
zhouzaida, line 1610: It would be convenient if `cfg` can be a file path. We can check if `cfg` is a string, calling `fromfile` to obtain the config.

HAOCHENYE:
BTW,do we need to sort the keys of the `Config` to ensure that the result of `pretty_text` is order-independent?

gachiemchiep:
> BTW,do we need to sort the keys of the `Config` to ensure that the result of `pretty_text` is order-independent?

Yes it is needed. 

> Performing list(d) on a dictionary returns a list of all the keys used in the dictionary, in **insertion order** (if you want it sorted, just use sorted(d) [in](https://docs.python.org/3.7/reference/expressions.html#in)stead). To check whether a single key is in the dictionary, use the in keyword.

https://docs.python.org/3.7/tutorial/datastructures.html#dictionaries

mmengine/config/config.py
gachiemchiep: Added in newest commit

gachiemchiep:
@HAOCHENYE  I added the sort so the output of pretty_text will be in order.

mmengine/config/config.py
```diff
@@ -1605,6 +1607,23 @@ def merge_from_dict(self,
...
+
+        if isinstance(cfg2, str):
+            cfg2_config = Config.fromfile(cfg2)
+        else:
+            cfg2_config = cfg2
+
+        res = difflib.unified_diff(
+            cfg1_config.pretty_text.split('\n'),
+            cfg2_config.pretty_text.split('\n'))
+        return '\n'.join(res)
```
zhouzaida, line 1625:
```suggestion
        if isinstance(cfg1, str):
            cfg1 = Config.fromfile(cfg1)

        if isinstance(cfg2, str):
            cfg2 = Config.fromfile(cfg2)

        res = difflib.unified_diff(
            cfg1.pretty_text.split('\n'),
            cfg2.pretty_text.split('\n'))
        return '\n'.join(res)
```

tests/data/config/py_config/test_diff_1.py
zhouzaida, line 1: ap

tests/data/config/py_config/test_diff_2.py
zhouzaida, line 1: same as above

HAOCHENYE:
Hi, I found that the difference processed by rich could look more pretty :smile: 

```python
from rich.console import Console
from rich.text import Text

...


console = Console()
text = Text()
for line in res:
    if line.startswith('+'):
        color = 'bright_green'
    elif line.startswith('-'):
        color = 'bright_red'
    else:
        color = 'bright_white'
    line = Text(line + '\n')
    line.stylize(color, 0, -2)
    text.append(line)

with console.capture() as capture:
    console.print(text)

return capture.get()
```

result:

image

gachiemchiep:
@HAOCHENYE 
I don't know why but after adding the rich.text, the pre-commit complained about some no attribute error. 

```bash
(openmmlab) gachiemchiep:mmengine$ pre-commit run --all-files 
flake8...................................................................Passed
isort....................................................................Passed
yapf.....................................................................Passed
trim trailing whitespace.................................................Passed
check yaml...............................................................Passed
fix end of files.........................................................Passed
fix requirements.txt.....................................................Passed
fix double quoted strings................................................Passed
check for merge conflicts................................................Passed
fix python encoding pragma...............................................Passed
mixed line ending........................................................Passed
mdformat.................................................................Passed
codespell................................................................Passed
docformatter.............................................................Passed
pyupgrade................................................................Passed
check copyright..........................................................Passed
remove improper eol in cn docs...........................................Passed
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

mmengine/config/config.py:1634: error: "str" has no attribute "stylize"
Found 1 error in 1 file (checked 244 source files)
```

I tried to change this to newer version but no luck 

```bash
# pre-commit-config.yaml
  - repo: https://github.com/pre-commit/mirrors-mypy
    rev: v0.812
    hooks:
      - id: mypy
        exclude: "docs"
```

mmengine/config/config.py
```diff
@@ -1605,6 +1609,36 @@ def merge_from_dict(self,
...
+        for line in res:
+            if line.startswith('+'):
+                color = 'bright_green'
+            elif line.startswith('-'):
+                color = 'bright_red'
+            else:
+                color = 'bright_white'
+            line = Text(line + '\n')
+            line.stylize(color)
+            text.append(line)
```
zhouzaida, line 1635:
```suggestion
            _text = Text(line + '\n')
            _text.stylize(color)
            text.append(_text)
```

zhouzaida: The above modifications may pass the lint.

mmengine/config/config.py
```diff
@@ -1432,7 +1435,8 @@ def _format_dict(input_dict, outest_level=False):
             use_mapping = _contain_invalid_identifier(input_dict)
             if use_mapping:
                 r += '{'
-            for idx, (k, v) in enumerate(input_dict.items()):
+            for idx, (k, v) in enumerate(
+                    sorted(input_dict.items(), key=lambda x: str(x))):
```
zhouzaida, line 1439:
```suggestion
                    sorted(input_dict.items(), key=lambda x: str(x[0]))):
```

zhouzaida: It may be enough to only sort it by the key (`x[0]`).

mmengine/config/config.py
gachiemchiep: Thanks. the first trick worked.

mmengine/config/config.py
gachiemchiep: I don't think so. 

configuration file's key in other project like mmdetection has the following naming format. 

```bash
...
train_pipeline = ...
train_dataloader = ....
val_pipeline = ...
val_dataloader = ....
val_evaluator = ... 
...
```
So 1 character won't be enough.
I think 4 characters ("train") is a better value.

zhouzaida: x[0] means the first item (corresponding to key) rather than the first char of key.

gachiemchiep: Oh i got it now.  so sleepy in the morning

zhouzaida: Thanks for your contribution.

zhouzaida (approved):

HAOCHENYE (approved):

PR #1266, opened 2023-07-21
Support multi-node distributed training with MLU backend

josh6688:
## Motivation

Fix the problem that multi-node distributed training would fail at initial when training with MLU.

## Modification

In `mmengine/dist/util.py`, we change rank to local rank when initializing process group with MLU.

## BC-breaking (Optional)

None

## Use cases (Optional)

None

## Checklist

### DIFF
--- a/mmengine/dist/utils.py
+++ b/mmengine/dist/utils.py
@@ -101,10 +101,11 @@ def _init_dist_pytorch(backend, init_backend='torch', **kwargs) -> None:
     rank = int(os.environ['RANK'])
     if is_mlu_available():
         import torch_mlu  # noqa: F401
-        torch.mlu.set_device(rank)
+        local_rank = int(os.environ['LOCAL_RANK'])
+        torch.mlu.set_device(local_rank)
         torch_dist.init_process_group(
             backend='cncl',
-            rank=rank,
+            rank=local_rank,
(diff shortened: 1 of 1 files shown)


mmengine/dist/utils.py
```diff
@@ -101,10 +101,11 @@ def _init_dist_pytorch(backend, init_backend='torch', **kwargs) -> None:
     rank = int(os.environ['RANK'])
     if is_mlu_available():
         import torch_mlu  # noqa: F401
-        torch.mlu.set_device(rank)
+        local_rank = int(os.environ['LOCAL_RANK'])
+        torch.mlu.set_device(local_rank)
         torch_dist.init_process_group(
             backend='cncl',
-            rank=rank,
+            rank=local_rank,
```
zhouzaida, line 108: Should the rank be a number between 0 and world_size-1 rather than the local rank?

josh6688: Sorry for keeping you waiting. The rank in init_process_group is changed to global rank. The code is tested with `example/distributed_training_with_flexible_runner.py`. Now it can run resnet50 with multi-node properly.

Danielmic (approved):

zhouzaida (approved):

PR #1300, opened 2023-08-10
[Feature] Add RecorderHook

Xinyu302:
## Motivation

The user expects to visualize the output of any layer of the network in the visualization hook, not necessarily the output of nn.module.forward, but also intermediate variables, and will inevitably need to invasively modify the code to update the output into the message hub. Then retrieve the value of the message hub in the visualization hook for visualization. Rather than intrusively modifying the code, we want to have a scheme that can non-intrusively obtain the output of any layer of the network. In addition, it also needs to provide the ability to get properties of a specified instance.

## Modification

In short, use ast module to modify the forward function of `runner.model`.
Here is a simple example.

```python
class MMResNet50(BaseModel):

    def __init__(self):
        super().__init__()
        self.resnet = torchvision.models.resnet50()

    def forward(self, imgs, labels, mode):
        x = self.resnet(imgs)
        if mode == 'loss':
            return {'loss': F.cross_entropy(x, labels)}
        elif mode == 'predict':
            return x, labels
```

The API of RecorderHook is like this:
```python
cfg.custom_hooks = [
    dict(
        type='RecorderHook',
        recorders=[
            dict(type='FunctionRecorder', ...),  # Function recorder1
            dict(type='FunctionRecorder', ...),  # Function recorder2
            dict(type='AttributeRecorder', ...)  # AttributeRecorder1
            ... 
        ],
    )
]
```
RecorderHook uses FunctionRecorder and AttributeRecorder to record different things in `forward` method.

#### function

Gets the output and intermediate variables of the specified function or method.

#### case

```python
    custom_hooks=[
        dict(
            type='RecorderHook',
            recorders=[dict(type='FunctionRecorder', target='x')],
            save_dir='./work_dir',
            print_modification=True)
    ],

```

#### Forward method after modification

```python
def forward(self, imgs, labels, mode):
    from mmengine.logging import MessageHub
    message_hub = MessageHub.get_current_instance()
    x = self.resnet(imgs)
    message_hub.update_info('x', x)
    if mode == 'loss':
        return {'loss': F.cross_entropy(x, labels)}
    elif mode == 'predict':
        return (x, labels)

```

#### function

Gets the value of the specified property.

#### case

```python
    custom_hooks=[
        dict(
            type='RecorderHook',
            recorders=[dict(type='AttributeRecorder', target='self.resnet', attribute='weight')],
            save_dir='./work_dir',
            print_modification=True)
    ],

```

#### Forward method after modification
*There is bug here*

```python
def forward(self, imgs, labels, mode):
    from mmengine.logging import MessageHub
    message_hub = MessageHub.get_current_instance()
    self_resnet_weight = self.resnet(imgs)
    message_hub.update_info('self_resnet_weight', self_resnet_weight.weight)
    x = self_resnet
    if mode == 'loss':
        return {'loss': F.cross_entropy(x, labels)}
    elif mode == 'predict':
        return (x, labels)
```

## TODO

* Add test cases.
* Recorder attribute of child module. For example, record tensors in self.resnet's forward method.
* Tensors could have the same name. Record all of them or should the user points out which one to record?

### DIFF
--- a/mmengine/hooks/recorder_hook.py
+++ b/mmengine/hooks/recorder_hook.py
@@ -5,10 +5,53 @@ import inspect
 import textwrap
 import types
 from typing import Any, List, Optional, Tuple, Union
+from collections import defaultdict
 
 from mmengine.logging import HistoryBuffer, MessageHub
 from mmengine.registry import HOOKS
 from . import Hook
+from abc import ABCMeta, abstractmethod
+
+
+class Recorder(metaclass=ABCMeta):
+    def __init__(self, target: str):
+        self._target = target
+
+    @abstractmethod
+    def rewrite(self, ast_tree):
+        pass
+
+# FunctionRecorder
+class FunctionRecorder(Recorder):
+    def __init__(self, target: str):
+        super().__init__(target)
+        self.visit_assign = self._get_adder_class()
+    # super.__init__()
+
+    def _get_adder_class(self):
+        outer_class = self
+        class FunctionRecorderAdder(ast.NodeTransformer):
+            def visit_Assign(self, node):
+                if node.targets[0].id != outer_class._target:
+                    return node
+                add2messagehub = ast.Expr(
+                    value=ast.Call(
+                        func=ast.Attribute(
+                            value=ast.Name(id='message_hub', ctx=ast.Load()),
+                            attr='update_info',
+                            ctx=ast.Load()),
+                        args=[
+                            ast.Constant(value = node.targets[0].id),
+                            ast.Name(id=node.targets[0].id, ctx=ast.Load())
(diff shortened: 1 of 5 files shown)


mmengine/hooks/recorder_hook.py
```diff
@@ -0,0 +1,396 @@
...
+
+import torch
+
+from mmengine.logging import MessageHub, print_log
+from mmengine.registry import HOOKS, RECORDERS
+from . import Hook
+
+
+class FunctionRecorderTransformer(ast.NodeTransformer):
+
```
zhouzaida, line 20: Please add docstring and type hint.

PR #1347, opened 2023-09-08
[CodeCamp2023-668] Add `AimVisBackend`

### LINKED ISSUE #1115: Proposal for Aim Visualizer Backend
### What is the feature?

Aim is a great experiment tracker. 

Currently no visualizer implemented for it. 

Not thinking the below is complete (I use the datetime to create the experiment name), but may be useful for other users. Happy to make changes to actually get it merged in. 

```
import aim
import datetime

@VISBACKENDS.register_module()
class AimVisBackend(BaseVisBackend):
    """
    Based on https://github.com/open-mmlab/mmengine/blob/main/mmengine/visualization/vis_backend.py 
    """

    def __init__(
            self,
            save_dir: str,  
        ):
        super().__init__(save_dir)
        self._init_env()

        
    def _init_env(self):
        """Setup env for MLflow."""

        try:
            import aim
        except ImportError:
            raise ImportError(
                'Please run "pip install aim" to install aim'
            )  # type: ignore
            
        clean_datetime = str(datetime.now()).replace(' ','_').replace(':','-')
        self._aim_run = aim.Run(
            experiment = clean_datetime,
        )
        
    @property  # type: ignore
    @force_init_env
    def experiment(self):
        """Return MLflow object."""
        return self._aim_run
    
    @force_init_env
    def add_config(self, config, **kwargs) -> None:
        """Record the config to mlflow.
        Args:
            config (Config): The Config object
        """
        # works perfectly
        self._aim_run['hparams'] = dict(config)
        
        # looks a bit clumsy in Aim UI
        aim_text = aim.Text(str(config))
        self._aim_run.track(aim_text, name='Config', step=0)
        
    @force_init_env
    def add_scalar(self,
                   name: str,
                   value: Union[int, float, torch.Tensor, np.ndarray],
                   step: int = 0,
                   **kwargs) -> None:
        """Record the scalar data to mlflow.
        Args:
            name (str): The scalar identifier.
            value (int, float, torch.Tensor, np.ndarray): Value to save.
            step (int): Global step value to record. Default to 0.
        """
        self._aim_run.track(name=name, value=value, step=step)
        
        
    @force_init_env
    def add_scalars(self,
                    scalar_dict: dict,
                    step: int = 0,
                    file_path: Optional[str] = None,
                    **kwargs) -> None:
        """Record the scalar's data to wandb.
        Args:
            scalar_dict (dict): Key-value pair storing the tag and
                corresponding values.
            step (int): Useless parameter. Wandb does not
                need this parameter. Defaults to 0.
            file_path (str, optional): Useless parameter. Just for
                interface unification. Defaults to None.
        """
        for k in scalar_dict:
            self._aim_run.track(name=k, value=scalar_dict[k], step=step)
```

### Any other context?

The use of 'save_dir' as the kwarg fed through to the VisualizerBackend seems a little odd, 
feels like it comes from the Local Visualizer but has no analogy for most other forms, e.g. MLFLow, Neptune, W and B. 

```
TypeError: class `DetLocalVisualizer` in mmdet/visualization/local_visualizer.py: class `AimVisBackend` in viz_backend.py: __init__() got an unexpected keyword argument 'save_dir'
```

Surely a better core would be experiment_name or experiment_id?

RangeKing:
## Motivation

Add `AimVisBackend`

resolve #1115

### DIFF
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -1355,10 +1355,11 @@ class AimVisBackend(BaseVisBackend):
 
         from datetime import datetime
 
-        path_list = os.path.normpath(self._save_dir).split(os.sep)
-        exp_name = f'{path_list[-2]}_{path_list[-1]}' \
-            if self._save_dir is not None \
-            else datetime.now().strftime('%Y%m%d_%H%M%S')
+        if self._save_dir is not None:
+            path_list = os.path.normpath(self._save_dir).split(os.sep)
+            exp_name = f'{path_list[-2]}_{path_list[-1]}'
+        else:
+            exp_name = datetime.now().strftime('%Y%m%d_%H%M%S')
 
         if self._init_kwargs is None:
             self._init_kwargs = {}
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -1318,7 +1318,7 @@ class AimVisBackend(BaseVisBackend):
     Examples:
         >>> from mmengine.visualization import AimVisBackend
         >>> import numpy as np
-        >>> aim_vis_backend = AimVisBackend(save_dir='temp_dir')
+        >>> aim_vis_backend = AimVisBackend()
         >>> img=np.random.randint(0, 256, size=(10, 10, 3))
         >>> aim_vis_backend.add_image('img', img)
         >>> aim_vis_backend.add_scalar('mAP', 0.6)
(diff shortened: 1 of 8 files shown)


mmengine/visualization/vis_backend.py
```diff
@@ -1309,3 +1309,132 @@ def _to_dvc_paramlike(self,
...
+        return self._aim_run
+
+    @force_init_env
+    def add_config(self, config, **kwargs) -> None:
+        """Record the config to Aim.
+
+        Args:
+            config (Config): The Config object
+        """
+        self._aim_run['hparams'] = dict(config)
```
zhouzaida, line 1382:
```suggestion
        if isinstance(config, Config):
            config = config.to_dict()
        self._aim_run['hparams'] = config
```

zhouzaida (approved):

Issue #1081, opened 2023-04-17
[Feature] `Visualizer` do not need `master_only` when saving image


BIGWangYuDong:
### What is the feature?

In Visualizer, saving image will using `master_only` function, when unsing multi-gpus, it will only save the master gpu images. 

In my opinion, we can delete this (have tried and it works)

HAOCHENYE:
It makes sense to me, would mind creating a PR to remove the `master_only`?

PR #1353, opened 2023-09-11
 [CodeCamp2023-327]Control whether methods of the Visualizer are executed only in the main process through a parameter

DeliMm:
## Motivation

Control whether methods of the Visualizer are executed only in the main process through a parameter.. See more details in https://github.com/open-mmlab/mmengine/issues/1081

## Modification

To implement this function, we add a parameter 'img_only_master' to `visualizer` by all kinds of `vis_backend`. During runtime, it is determined whether to retain only the images in the GPU by checking if img_only_master is passed and examining its corresponding value. The modified `master_only` decorator is as following:

```python
def master_only(func: Callable) -> Callable:
    """Decorate those methods which should be executed in master process.

    Args:
        func (callable): Function to be decorated.

    Returns:
        callable: Return decorated function.
    """

    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        if ('img_only_master' in kwargs
                and kwargs['img_only_master'] is False) or is_main_process():
            return func(*args, **kwargs)

    return wrapper
```
When we build the runner, we can follow this way to pass the `img_only_master`:
```python 
visualizer=dict(¬
             type='Visualizer',
             vis_backends=[
                 dict(
                     type='LocalVisBackend',
                     save_dir='./vis',
                     img_only_master=False)
             ]),
```

When we use the `add_img` method, we need pass the `img_only_master` through `img_only_master=runner.visualizer.img_only_master` 
## Checklist


HAOCHENYE:
Coupling the implementation of `master_only` with the `Visualizer` is not a good practice. I believe we can avoid using the `master_only` decorator inside the `Visualizer` and instead opt for adding a `master_only` parameter to the `Visualizer`. Then, in various methods, we can check whether `master_only` is set to True and control whether to invoke the corresponding method only on rank 0.


```python
class Visualizer
    def __init__(self, ..., master_only=True):
        ...
        self.master_only = master_only

    def xxx(self, ...):
        if _should_do():
            ...
    
    def _should_do(self):
        if self.master_only and get_rank() != 0:
            return False
        else:
            return True
```

DeliMm:
Thank you for your response. I initially thought that I could only add 'master_only' without modifying the decorator. And I'll make the necessary adjustments as soon as possible.

mmengine/visualization/visualizer.py
```diff
@@ -215,19 +215,24 @@ def __init__(
         if image is not None:
             self.set_image(image)
 
+    def _should_do(self):
+        if self.master_only and get_rank() != 0:
+            return False
+        else:
+            return True
```
HAOCHENYE, line 222:
```suggestion
        return get_rank() == 0 or not self.master_only
```

mmengine/visualization/visualizer.py
```diff
@@ -215,19 +215,24 @@ def __init__(
...
     def dataset_meta(self) -> Optional[dict]:
         """Optional[dict]: Meta info of the dataset."""
         return self._dataset_meta
 
     @dataset_meta.setter  # type: ignore
-    @master_only
     def dataset_meta(self, dataset_meta: dict) -> None:
         """Set the dataset meta info to the Visualizer."""
-        self._dataset_meta = dataset_meta
+        if self._should_do():
```
HAOCHENYE, line 233: We need to reconsider that if it is necessary to apply `self._should_do` here. I think we could skip this check here.

HAOCHENYE:
Hi, please check the modification will not break the CI

mmengine/visualization/visualizer.py
DeliMm: For this, I added the check based on the original master_only decorator for this function.  For dataset_meta, it might be better to keep the original master_only decorator.

PR #1355, opened 2023-09-12
[Feature] Support pipeline parallelism model wrapper

fanqiNO1:
# Background

As the model inference process requires more and more CUDA memory, we need a way to complete the model inference process in a variety of CUDA memory situations, mainly the following two cases:
1. Insufficient CUDA memory
    The model inference process is accomplished by cpu offload, disk offload policy.
2. Sufficient CUDA memory
    The model can be partitioned across multiple gpus, in which case the model inference should be done as efficiently as possible.

`huggingface` introduces the `accelerate` library, which can also allow users to complete the inference in the case of insufficient CUDA memory, but its utilization of resources is too inefficient.

# Design

To accelerate the inference process by utilizing resources as much as possible, we will implement a pipeline parallelism-based model wrapper.

The pipeline parallelism-based model wrapper is primarily responsible for:
- build model, load and dispatch weights
- pipeline parallelism-based inference process

This PR will support `MMPipelineParallel`.

# Environment

- PyTorch: 2.0.0
- CUDA: 11.8
- GPU: 8 * A100, 80G

# Validation
[ ] init_device_map
[ ] pipeline parallelism

# Experiment

Work in progress.

### DIFF
--- a/mmengine/model/wrappers/pipeline_distributed.py
+++ b/mmengine/model/wrappers/pipeline_distributed.py
@@ -4,17 +4,17 @@ from contextlib import contextmanager
 from functools import partial
 from queue import Queue
 from threading import Event, Thread, current_thread
-from typing import Dict, List, Iterable, Union, Optional, Tuple
+from typing import Any, Dict, List, Union, Optional, Tuple
 
 import psutil
 import torch
 import torch.nn as nn
 
-from mmengine.analysis import get_model_complexity_info
+from mmengine.analysis import FlopAnalyzer
 from mmengine.config import Config, ConfigDict
-from mmengine.registry import MODEL_WRAPPERS
+from mmengine.registry import MODELS, MODEL_WRAPPERS
 
-ModelType = Union[Config, ConfigDict, dict, nn.Module, str]
+ModelType = Union[Config, ConfigDict, dict, nn.Module]
 
 
 @MODEL_WRAPPERS.register_module()
@@ -29,7 +29,7 @@ class MMPipelineParallel(nn.Module):
     hook_visited_times: Dict[str, int] = {}
 
     def __init__(self,
-                 model: Optional[ModelType] = None,
+                 model: ModelType,
                  weights: Optional[str] = None,
                  num_pipelines: Optional[int] = None,
                  num_chunks: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


mmengine/model/wrappers/pipeline_distributed.py
```diff
@@ -0,0 +1,1199 @@
...
+        # init offload directory
+        if offload_directory is not None:
+            self.offload_directory = offload_directory
+        else:
+            self.offload_directory = tempfile.mkdtemp()
+        self.offloaded_weights: Dict[str, Dict[str, Any]] = {}
+
+        # init queues
+        MMPipelineParallel.in_queues, MMPipelineParallel.out_queues = \
+            self._init_queues()
```
zhouzaida, line 136: Why do we not assign them to self?

```python
self.in_queues, self.out_queues =self._init_queues()
```

mmengine/model/wrappers/pipeline_distributed.py
```diff
@@ -0,0 +1,1199 @@
...
+        else:
+            self.num_chunks = self.num_pipelines * 32
+
+        # init memory map
+        self.memory_threshold = memory_threshold
+        self.memory_map = self._init_memory_map(memory_map)
+
+        # init device map
+        self.no_split_module_classes = [] if no_split_module_classes is None \
+            else no_split_module_classes
```
zhouzaida, line 110:
```suggestion
        self.no_split_module_classes = no_split_module_classes or []
```

mmengine/model/wrappers/pipeline_distributed.py
fanqiNO1: Since these variables may be called by other in-class functions, such as `MMPipelineParallelism.Hook`, calling `self.events` at `MMPipelineParallelism.Hook._exit_prev_part` is wrong. Therefore, in order to standardize the writing style, the in-class class (`Hook` and `Flag`), these variables and static functions should be called by `MMPipelineParallel.xxxx`.

mmengine/model/wrappers/pipeline_distributed.py
```diff
@@ -0,0 +1,1213 @@
...
+            modules_i = modules[i]
+            for module in modules_i['modules']:
+                device_map[module] = {
+                    'part_id': i,
+                    'init_device': modules_i['init'],
+                    'exec_device': modules_i['exec']
+                }
+        # handle tied weights
+        tied_weights = self.model_tree['tied_parameters']
+        for source, targets in tied_weights.items():
```
HAOCHENYE, line 479: The key of tied_weights means `param_name`, and value means list of `module_name`s. So, why do we use `device_map[source]` and `device_map[target]` here?

PR #1366, opened 2023-09-23
[feature] Support multi-gpu test multiple datasets separately

Mactarvish:


zhouzaida:
Hi @Mactarvish , thank you for your contribution. I think this feature can be implemented in a new Evaluator `MultiDatasetEvaluator` rather than in the `Metric`. We can discuss this first in this PR.

Mactarvish:
> Hi @Mactarvish , thank you for your contribution. I think this feature can be implemented in a new Evaluator `MultiDatasetEvaluator` rather than in the `Metric`. We can discuss this first in this PR.

Thanks for your suggestion.  I've been trying this way.

PR #1388, opened 2023-10-11
[Fix] fix a bug when module is missing in low version of bitsandbytes

Ben-Louis:

## Motivation

In [`register_bitsandbytes_optimizers`](https://github.com/open-mmlab/mmengine/blob/6c5eebb823e3c9381d63fd0cd1873ed1bd9ee9de/mmengine/optim/optimizer/builder.py#L132C14-L154), some optimizer modules are missing from a low version of `bitsandbytes`, resulting in a bug for users with lower versions. (e.g. https://github.com/open-mmlab/mmpose/issues/2747)

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/optim/optimizer/builder.py
+++ b/mmengine/optim/optimizer/builder.py
@@ -132,6 +132,10 @@ SOPHIA_OPTIMIZERS = register_sophia_optimizers()
 def register_bitsandbytes_optimizers() -> List[str]:
     """Register optimizers in ``bitsandbytes`` to the ``OPTIMIZERS`` registry.
 
+    In the `bitsandbytes` library, optimizers that have the same name as the
+    default optimizers in PyTorch are prefixed with ``bnb_``. For example,
+    ``bnb_Adagrad``.
+
     Returns:
         List[str]: A list of registered optimizers' name.
     """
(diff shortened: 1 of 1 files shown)


mmengine/optim/optimizer/builder.py
```diff
@@ -146,7 +146,7 @@ def register_bitsandbytes_optimizers() -> List[str]:
                 'PagedAdamW8bit', 'LAMB8bit', 'LARS8bit', 'RMSprop8bit',
                 'Lion8bit', 'PagedLion8bit', 'SGD8bit'
         ]:
-            _optim = getattr(bnb.optim, module_name)
```
HAOCHENYE, line 149: Register a None optimizer seems weird, maybe we should skip registering the optimizer if it does not exist

Ben-Louis: If `_optim` is `None`, the result of `inspect.isclass(_optim)` is `False` and the registering will be skipped
https://github.com/open-mmlab/mmengine/blob/2135d3703ebffad6097c78cc4ce04aac2124360d/mmengine/optim/optimizer/builder.py#L149-L153

HAOCHENYE: From the perspective of code readability, explicitly skipping registration is preferable to implicitly skipping it by passing in `None`.

BTW, I'm not familiar with the optimizer in bitsandbytes, why don't we refer to the implementation here

https://github.com/open-mmlab/mmengine/blob/2135d3703ebffad6097c78cc4ce04aac2124360d/mmengine/optim/optimizer/builder.py#L120

 to register the optimizer?

Ben-Louis: Me neither. I just attempted to fix the bug with minimal changes. I'm uncertain if modifying the builder function for `bitsandbytes` to resemble `Sophia`'s could cause any issues

zhouzaida (approved):

Issue #1410, opened 2023-10-30
[Bug] Unable to Inherit Configuration Files Across Repositories


ryan-minato:
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmengine/issues) and [Discussions](https://github.com/open-mmlab/mmengine/discussions) but cannot get the expected help.
- [X] The bug has not been fixed in the latest version(https://github.com/open-mmlab/mmengine).

### Environment

Any OS and PyTorch version

### Reproduces the problem - code sample

According to the documentation ([url](https://mmengine.readthedocs.io/en/latest/advanced_tutorials/config.html#inherit-configuration-files-across-repository:~:text=MMEngine%20Installation%20specification%2C)), creating a Python library based on mmdet's file structure should allow for inheritance using {package_name}::, but in practice, this approach is blocked.

> Config will parse mmdet:: to find mmdet package and inherits the specified configuration file. Actually, as long as the setup.py of the repository(package) conforms to MMEngine Installation specification, Config can use {package_name}:: to inherit the specific configuration file.

### Reproduces the problem - command or script

```python
from mmengine.config import Config
cfg = Config.fromfile('test.py')
```

test.py
```python
_base_=['mylib::datasets/example.py']
```

### Reproduces the problem - error message

``` Shell
AssertionError: mmengine dose not support to load mylib config.
```

### Additional information

The issue likely originates from ([url](https://github.com/open-mmlab/mmengine/blob/e43bbb5e03a412ea07f9ef6d0f118586082c2845/mmengine/config/utils.py#L134-L136)), where a hardcoded dictionary is checked. 

```python
...
MODULE2PACKAGE = {
    'mmcls': 'mmcls',
    'mmdet': 'mmdet',
...
    'mmagic': 'mmagic',
}
...
# L134C1-L136C38
assert package in MODULE2PACKAGE, (
    f'mmengine does not support to load {package} config.')
package = MODULE2PACKAGE[package]
```

To bypass this problem, either add a key-value pair `mylib:mylib` to the `MODULE2PACKAGE` constant in mmengine/config/utils.py, or remove the check and replace it with importlib-based imports.

PR #1411, opened 2023-10-30
fix: Inherit Configuration Files Across Repositories.

minato-ellie:
## Motivation

Fix bugs with inherited configuration files.  #1410 

## Modification

Changed the logic for reading configuration files to work with any python package that conforms to the "MMEngine Installation specification" as described in the documentation.

> Config will parse mmdet:: to find mmdet package and inherits the specified configuration file. Actually, as long as the setup.py of the repository ( package) conforms to MMEngine Installation specification, Config can use {package_name}:: to inherit the specific configuration file.

Modified
``` python
    if package in MODULE2PACKAGE:
        package = MODULE2PACKAGE[package]
    else:
        try:
            importlib.import_module(package)
        except ModuleNotFoundError:
            raise ValueError(f'Cannot find package `{package}`')
```

Original([url](https://github.com/open-mmlab/mmengine/blob/e43bbb5e03a412ea07f9ef6d0f118586082c2845/mmengine/config/utils.py#L134-L136C38))
``` python
    assert package in MODULE2PACKAGE, (
        f'mmengine does not support to load {package} config.')
    package = MODULE2PACKAGE[package]
```

## BC-breaking (Optional)

Not expected. However, it may result in python packages that do not comply with the MMEngine Installation specification being able to be loaded just as well, making the error occur at a much later time, rather than on import.

## Checklist

### DIFF
--- a/docs/en/notes/contributing.md
+++ b/docs/en/notes/contributing.md
@@ -89,16 +89,16 @@ git commit -m "xxx" --no-verify
 
 ### 3. Create a development branch
 
-After configuring the pre-commit, we should create a branch based on the master branch to develop the new feature or fix the bug. The proposed branch name is `username/pr_name`
+After configuring the pre-commit, we should create a branch based on the main branch to develop the new feature or fix the bug. The proposed branch name is `username/pr_name`
 
 ```shell
 git checkout -b yhc/refactor_contributing_doc
 ```
 
-In subsequent development, if the master branch of the local repository is behind the master branch of "upstream", we need to pull the upstream for synchronization, and then execute the above command:
+In subsequent development, if the main branch of the local repository is behind the main branch of "upstream", we need to pull the upstream for synchronization, and then execute the above command:
 
 ```shell
-git pull upstream master
+git pull upstream main
 ```
 
 ### 4. Commit the code and pass the unit test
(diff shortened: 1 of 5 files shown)


HAOCHENYE:
Thank you for your contribution! Before we proceed, kindly sign the CLA :smile:. Are you indicating that you've adhered completely to OpenMMLab's distribution standards, including the metafile and copying `configs` to `site_packages/.mim`? 

Since there are strict rules to access config by `mmdet::xxx`, I believe a more appropriate approach would be to provide an interface that registers "mylib" with `PACKAGES2MODULES` and then verifies the integrity of "mylib's" structure.

minato-ellie:
> Thank you for your contribution! Before we proceed, kindly sign the CLA 😄. Are you indicating that you've adhered completely to OpenMMLab's distribution standards, including the metafile and copying `configs` to `site_packages/.mim`?
> 
> Since there are strict rules to access config by `mmdet::xxx`, I believe a more appropriate approach would be to provide an interface that registers "mylib" with `PACKAGES2MODULES` and then verifies the integrity of "mylib's" structure.

I'm not sure if I understand correctly, but are you referring to "registers" as the REGISTRY ([url](https://mmengine.readthedocs.io/en/latest/advanced_tutorials/registry.html))? If so, based on the mechanism of the REGISTRY, it seems that I have to import 'mylib' once before it can be used. This seems to affect the convenience of cross-module inheritance.

I plan to create a plugin system using entry_point to allow third-party packages to register themselves in MODULE2PACKAGE. Do you think this is a good approach?

PR #1471, opened 2024-01-10
[Fix] Fix the resume of iteration

LZHgrla:

## Motivation

`IterBasedTrainLoop` is not correct for resume. The beginning data index is 0, instead of `self._iter`

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/runner/loops.py
+++ b/mmengine/runner/loops.py
@@ -271,6 +271,14 @@ class IterBasedTrainLoop(BaseLoop):
         # In iteration-based training loop, we treat the whole training process
         # as a big epoch and execute the corresponding hook.
         self.runner.call_hook('before_train_epoch')
+        if self._iter > 0:
+            print_log(
+                f'Advance dataloader {self._iter} steps to skip data '
+                'that has already been trained',
+                logger='current',
+                level=logging.WARNING)
+            for _ in range(self._iter):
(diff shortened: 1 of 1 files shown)


mmengine/runner/loops.py
```diff
@@ -271,6 +271,14 @@ def run(self) -> None:
...
         # as a big epoch and execute the corresponding hook.
         self.runner.call_hook('before_train_epoch')
+        if self._iter > 0:
+            print_log(
+                f'Advance dataloader {self._iter} steps to skip data '
+                'that has already been trained',
+                logger='current',
+                level=logging.WARNING)
+            for _ in range(self._iter):
+                next(self.dataloader_iterator)
```
zhouzaida, line 281: It's a legacy issue. To avoid BC breaking, we can add a parameter to control it.

LZHgrla: The default value of `self._iter` is 0, `dataloader_iterator` is infinite. So, I think it will not lead to the BC breaking?

But, maybe I might have overlooked something?

zhouzaida: This behavior is inherited from mmcv so we did not fix it in mmengine. After consideration, I think it's a good proposal to fix the legacy issue.

zhouzaida (approved):

PR #1531, opened 2024-04-18
[fix] Handle case ClearML already been initialized

mhtrinh:
Handle case where ClearML already been initialized ealier by the user.

## Modification

Check if clearml already been initialized and not try to re-initialize again as ClearML don't allow that. 

## Use cases

In some use case, user may want to initialized ClearML to track logs prior to running the Runner. Eg: logging all the prep steps.


mhtrinh:
Actually, Clearml is smart enough to handle this.

PR #1564, opened 2024-07-21
Several Improvements for the latest PyTorch Framework

MGAMZ:
## Motivation

Fix torch Future Warning: 
FutureWarning: `torch.cuda.amp.GradScaler(args...)` is deprecated. Please use `torch.amp.GradScaler('cuda', args...)` instead.

## Modification

Deprecate torch.cuda.amp and transfer to torch.amp, and partial it with `device='cuda'`. The modification will not introduce any influence on the following codes in this file


MGAMZ:
## Motivation
When using torch.compile, the 'disable' key compile could override the compile. When disable is set to True, there should be no prompt like 'The model has been compiled'.


## Modification
Add a simple check when compile is a dict, and check if there exists a key 'disable'.

MGAMZ:
The pull request contains bad implementations, I will refine to a better version soon.

PR #1589, opened 2024-10-22
[Fix] Remove init due to breaking change in pytest 8.2.0

tibor-reiss:
#Fixes 1575

Breaking change introduced in pytest 8.2.0, see e.g. https://docs.pytest.org/en/stable/changelog.html#breaking-changes

See also this: https://github.com/pytest-dev/pytest/issues/12289#issuecomment-2340687937


GaetanLepage:
I can confirm that this patch works great.

PR #1548, opened 2024-05-23
[Fix] Speed up "--resume"

chtzs:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers. By the way, if you're not familiar with how to use pre-commit to fix lint issues or add unit tests, please refer to [Contributing to OpenMMLab](https://mmengine.readthedocs.io/en/latest/notes/contributing.html).

## Motivation

In the current version, "--resume" will make engine loading data like regular training, but discard all of this data before reaching the specified iteration. Therefore, the time required for resuming will not be much faster than starting a new training session.

## Modification

Added a new method called `skip_iter` , which skips data without triggering data loading by calling the `_next_index` in the built-in iterator of DataLoader

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDetection or MMPretrain.
4. The documentation has been modified accordingly, like docstring or example tutorials.

Issue #1520, opened 2024-03-23
[Feature] Speed up the resume process of IterBased loop


YinAoXiong:
### What is the feature?

https://github.com/open-mmlab/mmengine/blob/2c4516c62294964065d058d98799402f50afdef6/mmengine/runner/loops.py#L281
现有的恢复方式会对dataloader 迭代 n 个step，当n较大时，速度会很慢，因为执行了实际的数据加载和处理逻辑。 是否有比较好的方式只迭代index，不执行实际的数据加载流程。
1. 一种可能的方式是和用户约定一个返回虚拟数据的数据集接口，在恢复时返回虚拟数据，
``
```python
class Dataset:

    def __getitem__(self, index):
        if self._skip_flag:
            return # Fake data
        # 处理数据
        return Real data

    def skip(self):
        self._skip_flag = True

    def resume(self):
        self._skip_flag = False
            if (
                hasattr(self.dataloader.dataset, "skip")
                and callable(self.dataloader.dataset.skip)
                and hasattr(self.dataloader.dataset, "resume")
                and callable(self.dataloader.dataset.resume)
            ):
                self.dataloader.dataset.skip()
                for _ in range(self._iter):
                    next(self.dataloader_iterator)
                self.dataloader.dataset.resume()
            else:
                for _ in range(self._iter):
                    next(self.dataloader_iterator)
```
2. 方式一还是需要用户进行配合，是否可以对dataloader进行操作从而无感知的快速跳过？
```python
                iter_batch_sampler = iter(self.dataloader.batch_sampler)
                for _ in range(self._iter):
                    next(iter_batch_sampler)
```
尝试直接迭代batch_sampler 在worker=0的时候是正常的，在多worker的时候恢复数据顺序出现错误。 像知道有没有什么比较好的解决方案

zhouzaida:
一个最小改动的方案是在迭代前 mock dataset 的`__getitem__`方法：

```python
    def run(self) -> None:
        """Launch training."""
        self.runner.call_hook('before_train')
        # In iteration-based training loop, we treat the whole training process
        # as a big epoch and execute the corresponding hook.
        self.runner.call_hook('before_train_epoch')
        if self._iter > 0:
            print_log(
                f'Advance dataloader {self._iter} steps to skip data '
                'that has already been trained',
                logger='current',
                level=logging.WARNING)
            # mock
            old_getitem = self.dataloader_iterator.dataset.__getitem__
[...]

PR #1608, opened 2024-11-24
[Fix] Fix IterBased Loop Training with Faster Resume

henryzhengr:

## Motivation

Previous, [PR#1471](https://github.com/open-mmlab/mmengine/pull/1471) resolves the issue of IterBased loop resuming at the correct iteration. However, its approach requires processing through data loading and preprocessing, which significantly slows down the resume process. This PR introduces a more efficient solution that minimizes overhead while maintaining correctness. Compared to solutions proposed in [PR#1548](https://github.com/open-mmlab/mmengine/pull/1548) and [PR#1520](https://github.com/open-mmlab/mmengine/pull/1520), this solution achieves faster resume with fewer code changes.

## Modification

This solution leverages itertools.islice to skip iterations in the dataloader without actually triggering the data loading and preprocessing steps, significantly improving the speed of the resume process.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmengine/runner/loops.py
+++ b/mmengine/runner/loops.py
@@ -3,6 +3,7 @@ import bisect
 import logging
 import time
 from typing import Dict, List, Optional, Sequence, Tuple, Union
+from itertools import islice
 
 import torch
 from torch.utils.data import DataLoader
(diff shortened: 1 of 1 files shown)


4o3F:
Please merge this to fix the extreme slow resume

hujh1994:
It's wrong. The `dataloader_iterator` is not itered actually.
You can verify it from the codes below:
```
from itertools import islice

class OrderedNumberIterator:
    def __init__(self, start, end, step=1):
        self.current = start
        self.end = end
        self.step = step

    def __iter__(self):
        return self

    def __next__(self):
        if self.current < self.end:
            result = self.current
            self.current += self.step
            return result
        raise StopIteration

iterator = OrderedNumberIterator(1, 10)
islice(iterator, 0, 5)
print(next(iterator))
```

PR #1639, opened 2025-02-21
add weights_only argument for torch load to support new version of torch

dianyo:
## Motivation

torch.load start raise error in newer version if not specify weights_only=False, where the default behavior is True.

## Modification
- Update yapf using newer version to avoid `lib2to3` not support issue after python3.10 ref: https://github.com/google/yapf/issues/993
- Add weights_only=False to all torch.load calling in the codebase

## Checklist

- [X] Pre-commit or other linting tools are used to fix the potential lint issues.
- [X]  The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [ ]  If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDetection or MMPretrain.
- [ ]  The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,8 +12,8 @@ repos:
     rev: 5.11.5
     hooks:
       - id: isort
-  - repo: https://github.com/pre-commit/mirrors-yapf
-    rev: v0.32.0
+  - repo: https://github.com/google/yapf
+    rev: v0.43.0
     hooks:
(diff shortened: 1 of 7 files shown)


dianyo:
looks the circle ci is broken, i can do pre-commit and make sure that lint all looks good. It might comes from manifest files issue.
Screenshot 2025-02-21 at 4 36 45 PM

PR #1665, opened 2025-10-17
Multiple Enhancement to Make MMENGINE walk further

MGAMZ:
## Motivation

The maintenance progress of mmengine is getting slower and slower at present. In order to enable this excellent architecture to progress further, I have introduced some upcoming compatibility improvements and some minor optimizations for it.

### Packing according to PyPA

The packaging using `setup.py` is deprecated and is due to be removed. Now PyPA suggests to use `pyproject.toml` as the packaging configuration. So refactoring them now.

### Upgrade yapf from `0.32.0` to `0.43.0`

The original yapf is not supported in python3.13 due to the missing of `lib2to3`. And in order to use the latest yapf version, the repo site need to be changed, so the pre-commit-config.yaml undergo a small modification. This should only bring minor behavior difference. 

### Upgrade numpy from `1.2x` to `2.2.x+`

The `numpy.compat` is deprecated, using `numpy.fft` in `tests/data/config/lazy_module_config/test_ast_transform.py` and `tests/test_config/test_lazy.py`.

### torch.compile configuration

The `torch.compile` receives a lot args. And in mmengine, all contents in `compile` from any config file will be transferred to `torch.compile`.
The mmengine checks several dependencies when user specifies `compile` in their config file, and making `hasattr(config, compile)` as the flag of enabling `torch.compile`.
However this is not exact. The compile config can include an arg `disable` to state if the compiler is actually working. So the mmengine needs to carefully determine if the user is actually enabling the `torch.compile`. This caused a small modification in `mmengine/_strategy/base.py`

### Message Hub error caught

The message hub receives critical Tensor input and displays them. In it's procedure, the `item` method is called when handling values (can be lr, loss, metrics, time, epoch, etc...) When there exists some invalid values, the message simply crashes the process as there is an assertion.

Considering the user is easy to print something bad to the message hub at ANY code line, a more robust way is to should a warning and returns the value and see if the directly returned value can be successfully processed by outer functions.

### Message Hub ignore `torch.compile`

The message hub's operation will inevitablely triger graph breaking during compile, so disabling it now.

### Support for Python-Like config file containing `model_warpper_constructor` or `model_wrapper`

The current parsing of constructor_cfg does not support the following case (python-style configuration file):

```python
optim_wrapper = dict(
    type = DeepSpeedOptimWrapper,
    optimizer = dict(type=AdamW, lr=lr, weight_decay=weight_decay),
    accumulative_counts = grad_accumulation,
[...]

PR #1673, opened 2025-10-26
[Fix] editable install in `python>=3.13`

MGAMZ:
This is a sub-PR of #1665.

This PR fixes a compatibility issue in `get_version()` caused by relying on `exec()` to modify the local namespace.

The current implementation executes a version file with `exec(...)` and then attempts to retrieve `__version__` from `locals()`. While this may appear to work in `Python ≤3.12`, it breaks in `Python 3.13+` due to changes introduced by [PEP 667](https://peps.python.org/pep-0667/), which standardizes execution context behavior and prevents `exec()` from mutating a function’s local scope.

To ensure correctness and forward compatibility, this PR explicitly passes a dedicated dictionary as the global namespace to `exec()`, and safely reads `__version__` from that namespace.

Noted that `setup.py` is deprecated and we should include `pyproject.toml` according to the latest [PyPA](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/), and [PEP 621](https://peps.python.org/pep-0621/).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,9 +27,10 @@ def choose_requirement(primary, secondary):
 
 
 def get_version():
+    version_dict = {}
     with open(version_file) as f:
-        exec(compile(f.read(), version_file, 'exec'))
-    return locals()['__version__']
+        exec(compile(f.read(), version_file, 'exec'), version_dict)
+    return version_dict['__version__']
 
 
 def parse_requirements(fname='requirements/runtime.txt', with_version=True):


MGAMZ:
@HAOCHENYE This one ready to be reviewed.

HAOCHENYE:
Leave this PR for review last. I'm considering replacing setuptools with [hatch](https://github.com/pypa/hatch) as mmengine's new build system.

PR #1694, opened 2026-08-30
Save visualizer images on all processes

Marchematics:
Save visualizer images on all processes.


Marchematics:
Updated the regression test to reset the rank mock after Visualizer construction. LocalVisBackend now keeps rank 0 filenames unchanged and appends `_rankN` for other ranks, avoiding multi-process overwrites.

Marchematics:
Also removed the duplicate `isfile` mock flagged by the lint job; this was preventing the full pre-commit run from starting.

PR #1695, opened 2026-08-30
Fix syncbn training state

### LINKED ISSUE #1624: [Bug] convert_sync_batchnorm missing 'training' attribute
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmengine/issues) and [Discussions](https://github.com/open-mmlab/mmengine/discussions) but cannot get the expected help.
- [X] The bug has not been fixed in the latest version(https://github.com/open-mmlab/mmengine).

### Environment

(Issue obvious from source)

### Reproduces the problem - code sample

(Issue obvious from source)

### Reproduces the problem - command or script

(Issue obvious from source)

### Reproduces the problem - error message

(Issue obvious from source)

### Additional information

In [torch.nn.modules.batchnorm.py](https://github.com/pytorch/pytorch/blob/69b883d7ace02cb84a713c3a73e804ace1158eae/torch/nn/modules/batchnorm.py#L875) the `SyncBatchNorm.convert_sync_batchnorm()` method copies over the `training` attribute like `module_output.training = module.training`.

The [mmengine version](https://github.com/open-mmlab/mmengine/blob/a8c74c346d2ef3e5501115529ba588accb5f2a03/mmengine/model/utils.py#L250) is missing this. However it is present in the `revert_sync_batchnorm()` method right above.

Not having this will cause a NCCL timeout when the BN layer is kept in eval mode for fine-tuning while the model is in training mode. I figured this out due to a similar issue/solution [here](https://github.com/tianweiy/CenterPoint/issues/224#issuecomment-986228097)

Marchematics:
## Motivation

Preserve the source module's training mode when converting `BatchNorm` to `SyncBatchNorm`. Without this, evaluation-mode layers are silently switched to training mode, which can cause incorrect fine-tuning behavior and NCCL timeouts. Fixes #1624.

## Modification

Copy `module.training` to the converted module and add a regression test for evaluation-mode conversion.

## BC-breaking

No.

## Checklist

1. Targeted lint checks pass.
2. Added `test_convert_sync_batchnorm_keeps_training_state`.
3. No downstream changes are required.
4. No documentation changes are needed.

### DIFF
--- a/mmengine/model/utils.py
+++ b/mmengine/model/utils.py
@@ -248,6 +248,7 @@ def convert_sync_batchnorm(module: nn.Module,
         module_output.running_mean = module.running_mean
         module_output.running_var = module.running_var
         module_output.num_batches_tracked = module.num_batches_tracked
+        module_output.training = module.training
         if hasattr(module, 'qconfig'):
             module_output.qconfig = module.qconfig
     for name, child in module.named_children():
--- a/tests/test_hooks/test_checkpoint_hook.py
+++ b/tests/test_hooks/test_checkpoint_hook.py
@@ -336,7 +336,6 @@ class TestCheckpointHook(RunnerTestCase):
         petrel_client = MagicMock()
         for by_epoch, cfg in [(True, self.epoch_based_cfg),
                               (False, self.iter_based_cfg)]:
-            isfile = MagicMock(return_value=True)
             self.clear_work_dir()
             with patch.dict(sys.modules, {'petrel_client': petrel_client}), \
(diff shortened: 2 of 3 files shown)


Marchematics:
Removed the duplicate `isfile` mock that caused the lint failure in the checkpoint test. The changed test now passes the repository lint checks locally.

Marchematics:
Targeted regression test passes: `tests/test_model/test_convert_sync_batchnorm_training.py`.

PR #1696, opened 2026-08-30
fix checkpoint hook for standalone validation

Marchematics:
Initialize checkpoint state before standalone validation so save_best can write the best checkpoint.


Marchematics:
The branch already includes the duplicate-mock lint fix. The remaining CircleCI result is from the existing pre-commit environment; no additional code issue was found locally.

Marchematics:
Targeted checkpoint hook suite passes: 8 tests.

PR #1698, opened 2026-09-08
[Fix] Use SQLite for default MLflow tracking

### LINKED ISSUE #1687: [Bug] MlflowVisBackend fails due to use of deprecated filesystem store
### Prerequisite

- [x] I have searched [Issues](https://github.com/open-mmlab/mmengine/issues) and [Discussions](https://github.com/open-mmlab/mmengine/discussions) but cannot get the expected help.
- [x] The bug has not been fixed in the latest version(https://github.com/open-mmlab/mmengine).

### Environment

Updating `mlflow` to 3.14 and rebuilding its dependents triggers
> FAILED tests/test_visualizer/test_vis_backend.py::TestMLflowVisBackend::test_close - mlflow.exceptions.MlflowException: The filesystem tracking backend (e.g., './mlruns') is in maintenance mode and will not receive further updates. Please migrate to a database backend (e.g., 'sqlite:///mlflow.db') to access the latest MLflow features. The mlflow migrate-filestore tool migrates your existing data losslessly. See https://mlflow.org/docs/latest/self-hosting/migrate-from-file-store for migration guidance. If the filesystem backend is required for your workflow, set MLFLOW_ALLOW_FILE_STORE=true to opt out of this exception.

### Reproduces the problem - code sample

n/a

### Reproduces the problem - command or script

`nix-build -A python3Packages.mmengine` after updating `mlflow`.

### Reproduces the problem - error message

see above

### Additional information

_No response_

appleweiping:
## Motivation

MLflow 3.14 rejects filesystem tracking stores unless the compatibility escape hatch is enabled. This makes the default MLflowVisBackend initialization fail. Fixes #1687.

## Modification

When `tracking_uri` is not supplied, MLflowVisBackend now stores metadata in a SQLite database at `save_dir/mlflow.db` and artifacts under `save_dir/artifacts`. Explicit `tracking_uri` and `artifact_location` values keep their existing behavior.

The regression test verifies that the default tracking URI is SQLite and that the database is created. It uses a non-persistent SQLAlchemy pool during tests so Windows can remove the temporary database.

## BC-breaking

The default local metadata store changes from the legacy MLflow filesystem format to SQLite. Existing filesystem runs are not migrated automatically. They can be migrated with MLflow's `migrate-filestore` command. Explicit tracking URIs are unchanged.

## Validation

- Reproduced the failure with MLflow 3.14.0
- MLflowVisBackend tests: 8 passed
- Visualization backend test file: 37 passed and 14 skipped before unrelated DVCLive tests failed because optional `dvclive` and `pygit2` packages were not installed
- flake8, isort, YAPF, docformatter, and pyupgrade passed for the changed files

### DIFF
--- a/mmengine/visualization/vis_backend.py
+++ b/mmengine/visualization/vis_backend.py
@@ -8,6 +8,7 @@ import platform
 import warnings
 from abc import ABCMeta, abstractmethod
 from collections.abc import MutableMapping
+from pathlib import Path
 from typing import Any, Callable, List, Optional, Sequence, Union
 
 import cv2
@@ -662,7 +663,8 @@ class MLflowVisBackend(BaseVisBackend):
             Defaults to None.
         params (dict, optional): The params to be added to the experiment.
             Defaults to None.
-        tracking_uri (str, optional): The tracking uri. Defaults to None.
+        tracking_uri (str, optional): The tracking uri. If None, a local
+            SQLite database is created in ``save_dir``. Defaults to None.
         artifact_suffix (Tuple[str] or str, optional): The artifact suffix.
             Defaults to ('.json', '.log', '.py', 'yaml').
         tracked_config_keys (dict, optional): The top level keys of config that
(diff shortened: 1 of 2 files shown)


appleweiping:
The CircleCI lint job fails during pre-commit initialization, before it checks the changed files. Its Python 3.7 environment resolves a pre-commit version that rejects the hook manifest stage named pre-commit. The changed files pass flake8, isort, and YAPF locally.

